PR #16, opened 2015-01-02
Feature/us788 validate login

andycanterbury:


silbermm:
We don't want to merge into master!

PR #18, opened 2015-01-05
Feature/us701 add skill to my profile

silbermm:


silbermm:
Fixed the icon for the trash

PR #38, opened 2015-01-28
US584 Request to join a team

silbermm:
*Added an opportunity angular module
*Now catching exceptions in the Translation layer method for adding a
Response
*Added an integration test to make sure a bad responseId triggered an
exception


MinistryPlatform.Translation/Services/OpportunityService.cs
```diff
@@ -35,15 +35,26 @@ public static void RespondToOpportunity(string token, int opportunityId, string
...
                 if (ex.Message == "Sequence contains more than one element")
                 {
-                    throw new MultipleRecordsException("Multiple Participant records found! Only one participant allowed per Contact.");
+                    throw new MultipleRecordsException(
+                        "Multiple Participant records found! Only one participant allowed per Contact.");
                 }
+                return false;
 
             }
+            catch (System.ServiceModel.FaultException ex)
```
IngageMaddoxT: why not a generic catch to replace both, if you are only returning false?

PR #41, opened 2015-01-30
Fix for DE48/DE50

saylormic:


IngageMaddoxT:
@saylormic thanks for cleaning up all Ids

PR #60, opened 2015-02-05
Feature/us904 register on website error handling

shankx50:
This is supposed to wrap up the story.


crds-angular/app/modules/email_field/email_field.html
```diff
@@ -1,11 +1,19 @@
 ﻿<ng-form name="email_field">
     <!--TODO refactor the validation logic below into a shorter method-->
-    <div class="form-group" ng-class="{'has-error':  
+    <div class="form-group" ng-class="{'has-error':
          email_field.email.$error.required && submitted  && email_field.email.$dirty ||
```
silbermm: i really think we need to pull this logic out into the controller, for now other reason because it's really ugly and hard to follow.

crds-angular/app/modules/password_field/password_field.html
```diff
@@ -1,9 +1,14 @@
 ﻿<ng-form name="password_field">
-    <div class="form-group" ng-class="{'has-error': password_field.password.$error.minlength}">
+    <div class="form-group" ng-class="{'has-error': password_field.password.$error.minlength||
```
silbermm: i really think we need to pull this logic out into the controller, for now other reason because it's really ugly and hard to follow.

PR #61, opened 2015-02-06
Defect/de66 contextual message for birthdate

silbermm:


IngageRitterS:
There is an extraneous debug statement in the code....no big deal...or perhaps it was intentionally left there?  :)  I know we will go back thru at some time and remove all of these.

PR #68, opened 2015-02-12
Person service.  calling CreateRecord if address doesn't exists, otherwi...

IngageMaddoxT:
...se calls update (new users don't have an address)

In MP removed Administrator role from testme account, that broke a few tests.
Added AppSettings lookup to base service, cleans up app settings lookup
Lookup Service, cleaned up page id lookup,  removed unused call to Attributes
CreatePageRecord test, removed delete, need a different way to cleanup


IngageMaddoxT:
@IngageRitterS please review before I forward to Dan.  thanks

PR #73, opened 2015-02-14
Feature/us945 enable waiting list

danrye:


danrye:
I'd rather move this logic to a stored procedure and call the stored proc with the trigger... The only way I'm thinking of doing that involves saving out all those values from INSERTED, I don't like that, can I pass INSERTED to the stored proc?

danrye:
Also, I'm not sure if this would work if the insert or update contained multiple rows...

IngageMaddoxT:
:+1:

PR #77, opened 2015-02-16
Feature/de79 error messaging not working when password left blank

shankx50:
@silbermm


silbermm:
@shankx50 can you merge development into your branch and then push again? The code is not able to be merged currently.

PR #93, opened 2015-03-03
DE86 Email validation broken on registration

silbermm:
The email api was changed to accept the userId in the url instead of
looking in the cookie. When this change occurred, we forgot to check the
userId for undefined in javascript which will happen if we are not
logged in (the registration page is accessed when a user is not logged
in).

To fix this, I added a check in the email validator directive that
checks it the Session.exists('userId') is undefined and returns 0 if it
is.


IngageMaddoxT:
:+1:

PR #101, opened 2015-03-06
US300: Removed background color from atrium display

IngageKrizJ:
Per Tate (https://rally1.rallydev.com/#/27593764268d/detail/userstory/24467701533/discussion), removing background.


andycanterbury:
:+1:

PR #103, opened 2015-03-11
US1097 - Add Gateway method to return family members for contact

andycanterbury:


Gateway/MinistryPlatform.Translation.Test/TranslationTests.cs
```diff
@@ -17,6 +18,18 @@ public class TranslationTests
         private const string NEW_PASSWORD = "changemeagain";
 
         [Test]
+        public void GetMyFamily()
+        {
+            var token = AuthenticationService.authenticate(USERNAME, PASSWORD);
+            //var token = AuthenticationService.authenticate("tmaddox@aol.com", "crds1234");
```
danrye: remove this comment (probably should prune it from history)

PR #106, opened 2015-03-18
Feature/us907 sign up for comm group

IngageKrizJ:
Finalize changes for signing up for community group


crossroads.net/app/routes.js
```diff
@@ -158,8 +155,22 @@
             loggedin: checkLoggedin
         }
     })
+    .state("community-groups-signup", {
+        url: "/sign-up/:urlsegment",
+        controller: "GroupSignupController",
```
silbermm: Can we convert this to using the ControllerAs syntax? 
```javascript
controller: "GroupSignupController as signup"
```
It makes for cleaner controllers because we can then get rid of `$scope` and use:
```javascript 
this.showContent = true;
this.showSuccess = false;
this.showFull = false;
```

crossroads.net/app/community_groups_signup/index.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var app = require("angular").module('crossroads');
+
+    app.controller('GroupSignupController', ['$scope',
```
silbermm: if we use controllerAs syntax, we can remove `$scope` from the DI list

crossroads.net/app/community_groups_signup/group_signup_controller.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+require('../services/group_service');
+(function () {
+    module.exports = function GroupSignupController($scope, $rootScope, Profile, Group, $log, $stateParams, Page) {
```
silbermm: if we use controllerAs syntax, we can remove `$scope` from the DI list

PR #117, opened 2015-03-19
fixed issues from comments from Brian on story

saylormic:
"Where?" link on routing number should display a modal with a check image (see m.crossroads.net). We should try and find an SVG graphic for this or create our own.

Let's disable zoom on the page. http://getbootstrap.com/css/#overview-mobile

Update line-height for the subheading class to 1.25.

There needs to be a <div class="row"> within the ng-scope class and outside our col-sm-8... classes. These are nested classes so they need a row class, seehttp://getbootstrap.com/css/#grid-nesting (screenshot https://www.dropbox.com/s/cm7b0wslli56nin/Screenshot%202015-03-18%2020.36.09.png?dl=0)


andycanterbury:
:+1:

PR #135, opened 2015-03-23
DE93: Possible fix for CORS issue

IngageKrizJ:
Implemented fix @IngageMaddoxT found.

### DIFF
--- a/crossroads.net/app/app.config.js
+++ b/crossroads.net/app/app.config.js
@@ -1,16 +1,15 @@
 'use strict()';
 (function(){
 
-  var getCookie = require('./utilities/cookies'); 
-  
-  var app = require("angular").module("crossroads"); 
+  var getCookie = require('./utilities/cookies');
+
+  var app = require("angular").module("crossroads");
   app.config(AppConfig);
-  
+
   function AppConfig($httpProvider){
     $httpProvider.defaults.useXDomain = true;
     $httpProvider.defaults.headers.common['Authorization']= getCookie('sessionId');
+    $httpProvider.defaults.headers.common['Access-Control-Allow-Methods']= 'GET';
   }
 
 })();
-
-


silbermm:
This doesn't make sense, but I am going to merge since it seems to work!

IngageKrizJ:
Thanks. I have an idea why this works, and will look at it tonight after my
kids baseball practice. It may actually be an angular bug, if it is what I
think it is.

IngageKrizJ:
Ok, I think I've found the issue.  What is happening is that for some
reason (not known), our CORS preflight OPTIONS request is going to the
server with header "Access-Control-Request-Headers: accept".  Per the W3C
spec on CORS (
http://www.w3.org/TR/2008/WD-access-control-20080912/#simple-request-header-whitelist),
Accept, Content-Type, and Accept-Language are "whitelisted" headers, which
do not need to be included in the "Access-Control-Request-Headers" request
header, and are not RETURNED in the Access-Control-Allowed-Headers response
header.  So the browser is sending accept up to the server as one of the
headers it wants to send, but it is not coming back on the preflight
response header, so it throws the "SEC7123: Request header accept was not
present in the Access-Control-Allow-Headers list." followed by the
"SCRIPT7002: XMLHttpRequest: Network Error 0x80070005, Access is denied."
error.

Additionally, sending ANY non-whitelisted header in the request will fix
this, as it will be added to the Access-Control-Request-Headers
(like Access-Control-Request-Headers: accept, x-use-the-force), and then
only the non-whitelisted header comes back in the
Access-Control-Allow-Headers response.

This seems like it is a bug in IE:
1) It should not be sending "accept" on this
preflight Access-Control-Request-Headers header in the first place, since
it is whitelisted.
2) It doesn't even seem to care if it gets the "accept" header back, seems
like it fails only because we don't get ANY response header when we only
send "accept"

I initially thought this might be an angular bug, but forgot that angular
really has nothing to do with the preflight request being sent, this is
built into the browser's XHR implementation (I think..)

So a couple better ways to fix this would be:

   - Send some other junk header in app.config.js
   - This works: $httpProvider.defaults.headers.common['X-Use-The-Force']=
      'true';
   - Explicitly set "Accept" header to empty in
   profile_unique_email_directive.js on the $http.get
      - This works: headers: { "Accept": "" }
   - Add some other junk header in profile_unique_email_directive.js on the
   $http.get
      - This works: headers: { "X-Use-The-Force":"true" }

Ideally, we could simply send back the "accept" header on the response, but
I can't seem to find a way to inject that into the response.  The
EnableCors attribute places some behavior around this, but there is no easy
way that I can tell to intercept this and add more headers (the
CorsMessageHandler writes the response).  In lieu of that, I think we ought
to switch to the third option for now.

On Mon, Mar 23, 2015 at 4:41 PM, Jim Kriz <jim.kriz@ingagepartners.com>
wrote:

> Thanks. I have an idea why this works, and will look at it tonight after
> my kids baseball practice. It may actually be an angular bug, if it is what
> I think it is.
> On Mar 23, 2015 4:07 PM, "Matt Silbernagel" <notifications@github.com>
> wrote:
>
>> This doesn't make sense, but I am going to merge since it seems to work!
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/crdschurch/crds-angular/pull/135#issuecomment-85171086>
>> .
>>
>


-- 
[image: Logo: Ingage Partners, Consulting with Purpose]
<http://www.ingagepartners.com/>

Jim Kriz
Senior Consultant
513.321.7700 (w) *|* 513.703.4218 (m)
2753 Observatory Ave *|* Cincinnati, OH 45208
www.IngagePartners.com <http://www.ingagepartners.com/>
[image: B-Corp Button] <http://www.bcorporation.net/ingagepartners> [image:
Linked-in Button] <http://www.linkedin.com/company/ingage-partners-llc> [image:
Facebook Button]
<http://www.facebook.com/pages/Ingage-Partners/337211479674920>

PR #151, opened 2015-03-30
US1138 Filter opportunities by time

silbermm:
Added `applyTimeFilter()` to `refineList` directive. Created method
called `filterAll()` which actually does the filtering when checking a
checkbox.


andycanterbury:
:+1:
:8ball:

PR #153, opened 2015-03-30
Feature/us1103 protect sensitive date

IngageMaddoxT:
My objective was to remove ApiUser and ApiPassword from common config.  

I'm storing the in ENV variables (I'll have to update the README as part of this story).  To read those I created a ConfigurationWrapper and interface for mocking.

As I was updating tests I was hoping I could add reading the web.config to my new wrapper.  I added that as well.  I updated as many tests, classes, services, controllers as I could justify in the scope of this story.  A few places are creating New instance of the wrapper.

@IngageKrizJ  @IngageRitterS would love your thoughts on this code as it exists now.  I'm still considering this "Developing" at this point.  I'm looking for a sanity check and suggestions.

thanks


IngageKrizJ:
Looks good so far.  I guess my only question is how are developers supposed to get the API user credentials when they start working?  Would we just have that documented somewhere (not in the README, since that is allso on github)?

As an aside, is there an way to encrypt values (not entire files, just values) in a config so that we could have the encrypted values still in the common.config?  We did this at my previous client, but it was in java Spring config files, not sure if we could do that here.

IngageMaddoxT:
There is a way to encrypt web.config/app.config section.  I tried it with
the common.config file, didn't go to well.  To be honest I didn't spend a
lot of time investigating that route.  It encrypted the appsettings section
of web.config, but left common config alone.  According to the
documentation it should be smart enough to traverse to the included file.

Since we have other ENV VARS for the project I was OK going that route.

And yes, we will have to document those values outside the README.

On Mon, Mar 30, 2015 at 3:41 PM, Jim Kriz <notifications@github.com> wrote:

> Looks good so far. I guess my only question is how are developers supposed
> to get the API user credentials when they start working? Would we just have
> that documented somewhere (not in the README, since that is allso on
> github)?
>
> As an aside, is there an way to encrypt values (not entire files, just
> values) in a config so that we could have the encrypted values still in the
> common.config? We did this at my previous client, but it was in java Spring
> config files, not sure if we could do that here.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/crdschurch/crds-angular/pull/153#issuecomment-87802954>
> .
>



-- 


[image: Logo: Ingage Partners, Consulting with Purpose]
<http://www.ingagepartners.com/>

Tony Maddox
Senior Consultant
859-803-2490 (m)
2753 Observatory Ave *|* Cincinnati, OH 45208
www.IngagePartners.com <http://www.ingagepartners.com/>
[image: B-Corp Button] <http://www.bcorporation.net/ingagepartners> [image:
Linked-in Button] <http://www.linkedin.com/company/ingage-partners-llc> [image:
Facebook Button]
<http://www.facebook.com/pages/Ingage-Partners/337211479674920>

Gateway/crds-angular/Controllers/API/AccountController.cs
```diff
@@ -27,7 +33,7 @@ public IHttpActionResult Get()
             {
                 try
                 {
-                    AccountService accountService = new AccountService();
+                    AccountService accountService = new AccountService(_configurationWrapper);
```
IngageKrizJ: Any reason this is still instantiating an AccountService, as opposed to AccountController being injected with one?  I assume it was just a matter of how much refactoring you could do on this story, I know there are a lot of dependencies...

IngageMaddoxT: You are correct.  I left some for later, because it was getting to be too much for this story.

PR #159, opened 2015-03-31
US1134 Added Breadcrumb Pattern

bzmillerboy:


tatelucas:
This looks good to me. The only tweak I might make is switch the .active breadcrumb from orange to our normal text color. The orange makes me think I can/should click on it but this is the only one they can't click.

PR #161, opened 2015-03-31
DE123: Update Contact Service to return nulls for optional nullable fields

IngageKrizJ:
Gender_ID and Marital_Status_ID are optional integer profile fields, but they were being returned as 0 instead of null when they were not set in the db.  This was causing issues when saving the profile, if those fields were not changed on the frontend.


andycanterbury:
:100:

PR #173, opened 2015-04-06
Feature/us1164 edit profile from serve signup

IngageMaddoxT:


crossroads.net/e2e/myserve/editProfile.js
```diff
@@ -0,0 +1,64 @@
+var env = require("../environment");
```
andycanterbury: For the other Jasmine tests we have used .spec.js in the file name. Do we want to come up with a similar standard to identify protractor/selenium test files?

crossroads0.net/e2e/myserve/displayMyFamilysOpportunities
```diff
@@ -0,0 +1,65 @@
+var env = require("../environment");
```
andycanterbury: Should this file have an extension of some kind?

crossroads.net/app/utilities/check_login.js
```diff
@@ -1,8 +0,0 @@
-'use strict()';
```
andycanterbury: Is this just cleaning up unused code?

andycanterbury: Discussed with @silbermm and this was unused code being cleaned up.

PR #174, opened 2015-04-08
Resolving defect - casting to Byte rather than string.  What was I think...

IngageRitterS:
...ing in the first place??


IngageBrewerM:
:+1:

PR #176, opened 2015-04-08
Defect/de132 cannot sign up family member first

shankx50:
This should fix the bug I encountered during demo.


IngageBrewerM:
Controller needs clean up, but the code removal is fixes the defect so :+1:

PR #181, opened 2015-04-10
Feature/us996 one time giving start page

shankx50:


crossroads.net/app/give/donationDetails.directive.js
```diff
@@ -0,0 +1,21 @@
...
+          templateUrl: 'give/donation-details.html',
+          restrict: 'EA'
+      };
+      return directive;
+
+      function link(scope, element, attrs) {
+        scope.programs = getPrograms.fetchPrograms();
+      }
+    }
+})()
```
IngageKrizJ: Should this directive reference the give controller, and then attach the programs list to the give controller instead of scope?

PR #184, opened 2015-04-10
Feature/us1201 one time giving bank info

IngageRitterS:


crossroads.net/app/give/bankAccountNumber.directive.js
```diff
@@ -0,0 +1,19 @@
...
+(function () {
+
+  module.exports = function () {
+
+        return {
+            restrict: "A",
+            require: 'ngModel',
+            link: function(scope, element, attrs, ngModel){
+                ngModel.$validators.invalidAccount = function (value) {
+                   var validAccount = /^[1-30]\d*$/;
```
IngageKrizJ: Is this supposed to match 1-30 numerics?  If so, I think that pattern needs to be something like "^[0-9]{1,30}$"

crossroads.net/app/give/routingTransitNumber.directive.js
```diff
@@ -0,0 +1,21 @@
...
+(function () {
+
+  module.exports = function () {
+
+        return {
+            restrict: "A",
+            require: 'ngModel',
+            link: function(scope, element, attrs, ngModel){
+                ngModel.$validators.invalidRouting = function (value) {
+                   var validRtn = /^[1-9]\d*$/;
```
IngageKrizJ: "^[0-9]{9}$" would match

PR #193, opened 2015-04-15
US1049 show success message when signing up

silbermm:
Added a key to the messages constant for the correct message in the
CMS and called sent a message with $rootScope.$emit to send the message.

Tried to write an e2e test to validate the correct message was showing,
but the growl library uses a $timeout which makes it impossible to test. See [https://github.com/JanStevens/angular-growl-2/issues/81](github) for more info.


IngageMaddoxT:
merged & deleted

PR #196, opened 2015-04-16
Feature/us1114 icon indicator

silbermm:
We created a directive to show the correct icon on the persons tab.


IngageBrewerM:
Code looks good, so :+1: to merge.  Wondering if there can be any protractor tests written around this?

PR #197, opened 2015-04-16
Feature/us1184

saylormic:


crossroads.net/app/mytrips/mytrips.html
```diff
@@ -0,0 +1,245 @@
+<div class="col-sm-10 col-sm-offset-1 my-trip">
+    <div class="row">
+        <h1>My Trips</h1>
+        <h2 class="subheading">Whether it’s New Orleans or India, we leave our own neighborhoods to be in others.</h2>
+
+
+
+        <div class="col-md-12 row">
```
bzmillerboy: You can't add a col and row class to the same element, this will cause all kinds of problems. Place "row" in it's own div.

crossroads.net/app/mytrips/mytrips.html
```diff
@@ -0,0 +1,245 @@
...
+            <div class="card-item" data-category="view">
+                <!-- card heading -->
+                <div class="card-header card-background-dark clearfix">
+                    <div class="col-sm-10 col-xs-8">
```
bzmillerboy: This is a nested col but there is no parent row, you should add a row when nesting col.

crossroads.net/app/mytrips/mytrips.html
```diff
@@ -0,0 +1,245 @@
...
+                        <p class="small text-muted trip-date">July 7 - July 27, 2015</p>
+                    </div>
+                    <div class="col-sm-2 col-xs-4 text-center hard-right days-left">
+                        <div class="card-row card-heading text-danger">35</div>
+                        <p>days left</p>
+                    </div>
+                </div>
+                <!-- /card heading -->
+                <div class="col-xs-12 col-top">
+                    <div class="card-header">
```
bzmillerboy: This probably needs to be a nested col. It's not incasing all of it's elements which is hiding the "Hide/Show" button.

PR #229, opened 2015-04-27
Feature/us1209 one time give new registration

IngageRitterS:
Can someone please review - @silbermm @andycanterbury @IngageMaddoxT @shankx50


crossroads.net/app/give/account.html
```diff
@@ -6,7 +6,47 @@
...
+	    <div id="existingEmail" growl inline="true" reference="{{give.emailAlreadyRegisteredGrowlDivRef}}" limit-messages="1"></div>
+
+  		<div class="form-group" >
+		    <div class="btn-group btn-group-sm account-type form-group" role="group">
+		        <button type="button" class="btn btn-ghost" ng-model="give.view" btn-radio="'bank'">Bank Account</button>
+		        <button type="button" class="btn btn-ghost" ng-model="give.view" btn-radio="'cc'">Credit Card</button>
+		    </div>
+
+		    <span>
+		        <a href="#/confidentiality-and-security/" target="_blank">
```
silbermm: can we use ui-sref here? this way, if we ever change the url (in the routes config) this link won't break

crossroads.net/app/give/give_controller.js
```diff
@@ -90,12 +175,17 @@
             }
         }
 
+        vm.routingError = function() {
+            return (vm.bankinfoSubmitted && $scope.giveForm.accountForm.routing.$error.invalidRouting && $scope.giveForm.accountForm.$invalid  ||
+                $scope.giveForm.accountForm.routing.$error.invalidRouting && $scope.giveForm.accountForm.routing.$dirty)
+        };
+
         vm.submitBankInfo = function() {
```
silbermm: can we add a comment to the file, or each function that changes state, that it first looks at the stateChangeStart and prevents changing state if there is an error. I remember looking at this on Friday with @IngageBrewerM and we all didn't know how the state change got rejected.

crossroads.net/app/give/login.html
```diff
@@ -1,15 +1,15 @@
 <div class="btn-group btn-group-justified" role="group">
-    <a type="button" class="btn btn-ghost active" role="button">Login</a>
+    <a type="button" class="btn btn-ghost active"  role="button">Login</a>
     <a type="button" class="btn btn-ghost" role="button" ui-sref="give.account">Give as Guest</a>
 </div>
 
 <h1>Login</h1>
 <p class="subheading">Login to have your payment information saved and receive a quarterly statement. If you don't already have an account,
-    <a href="" ui-sref="give.register">create an account</a>.
+    <a href="#/register" >create an account</a>.
```
silbermm: This should use ui-sref

silbermm:
*In GiveController I noticed that Profile is being injected and not being used. 
*To do injections we have switched to using the $inject method highlighted here https://github.com/crdschurch/angular-styleguide#manually-identify-dependencies
*getCookies does not appear to be needed in GiveController
*The give module is just named `give` . I think it's best to call it `crossroads.give` because give is such a common word that it could possible cause collisions. This should be true of all of our modules. 
*the file names for the validation directives are not named that same as the directive i.e(invalidZip is in a file called zipCode.validation.directive) this is rather confusing. The filename should match the directive name i.e (invalidZip.validation.directive)
*Do the creditCardInfo, bankInfo, donationConfirmation, donationDetails need to have their own module respectively? Can't they all live under the give module? If they remain their own modules, I recommend calling them `crossroads.give.ccInfo`, `crossroads.give.bankInfo`, `crossroads.give.donationConfirmation` and `crossroads.give.donationDetails`

PR #231, opened 2015-04-28
Participant Service: restricted search for participant by contact id to ...

IngageMaddoxT:
...first column of results.

Moved Get Participant call high in call chain to reduce duplicate calls.


Gateway/MinistryPlatform.Translation.Test/Services/ParticipantServiceTest.cs
```diff
@@ -30,7 +30,7 @@ public void GetParticipantByParticipantId()
             const int contactId = 99999;
 
             const string viewKey = "ParticipantByContactId";
-            var searchString = contactId.ToString();
+            var searchString = contactId.ToString() + ",";
```
silbermm: if you concat something to contactId with a `+`, do you still need to explicitly call toString()?

PR #233, opened 2015-04-28
Feature/us1244 retrieve available programs

shankx50:
@mattbrewer


IngageRitterS:
These changes should be made across the board for all other directives within 'give'

PR #234, opened 2015-04-29
Feature/us1231 migrate sql

danrye:


CI/updateSQLScriptEntries.ps1
```diff
@@ -0,0 +1,31 @@
...
+		$exitCode = $LASTEXITCODE
+	}
+  	$output = & sqlcmd -S mp-int-db.cloudapp.net -b -Q "Update [MinistryPlatform].[dbo].[cr_Scripts] set executed=1 where [MD5] = '$hash'"
+	if($LASTEXITCODE -ne 0){
+		echo "Error: $output"
+		$exitCode = $LASTEXITCODE
+	}
+  }
+}
+
```
IngageKrizJ: So if we have scripts like 1.sql, 2.sql, and 3.sql (in that order), and assume 1.sql works, 2.sql has errors, and 3.sql works, won't this exit with a successful exit code, and build will be green?  Even though 2.sql failed, the last exit code (from 3.sql) will be success.  Unless I'm missing something...

danrye: I don't believe so, because we only init exit code to 0, so if it ever gets set, it's getting set to 1... all the other sets are in a if($LASTEXITCODE -ne 0){

IngageKrizJ: Ah, I see that now - sorry, it's late, and my mind is going to mush looking at SSRS...

PR #235, opened 2015-04-29
Feature/us1140 filter rsvp

IngageMaddoxT:
@silbermm


Gateway/crds-angular/Controllers/API/ServeController.cs
```diff
@@ -39,6 +39,7 @@ public IHttpActionResult GetFamilyServeDays()
                 try
                 {
                     var servingDays = _serveService.GetServingDays(token);
+                    //var servingDays = MockResponseForFrontEndTesting();
```
silbermm: comment left in the code?

Gateway/crds-angular/Controllers/API/ServeController.cs
```diff
@@ -95,5 +96,243 @@ public IHttpActionResult SaveRsvp([FromBody] SaveRsvpDto saveRsvp)
                 return this.Ok();
             });
         }
+
+
+
+        //for testing only
+        private List<ServingDay> MockResponseForFrontEndTesting()
```
silbermm: is this still being used?

silbermm:
I think the JS looks good and there are the correct tests, but do you still need all of the fake data in the controller still?

PR #251, opened 2015-04-30
Feature/us883 deploy config

IngageMaddoxT:


Gateway/common.config
```diff
@@ -1,6 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <appSettings>  
   <!--Below are codes reflecting the respective pages on MP-->
+  <add key="test2" value="2"/>
+  <add key="test3" value="3"/>
+  <add key="test4" value="6"/>
```
andycanterbury: Do these need to stay in the config or can the be removed before merging?

Gateway/crds-angular.sln
```diff
@@ -20,11 +20,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MinistryPlatform.Translatio
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MinistryPlatform.Models", "MinistryPlatform.Models\MinistryPlatform.Models.csproj", "{CD41507E-91AC-4C60-96E2-5CAC25F49A76}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{7E3D7A92-C17E-4F71-AA49-6905A09FAE3E}"
-	ProjectSection(SolutionItems) = preProject
-		common.config = common.config
```
andycanterbury: This removes common.config from the solution but does it also need to be removed the repo? No point in keeping an unused file?

PR #258, opened 2015-05-01
Defect/de147 donation details refactor

shankx50:
The main goal of this defect is to make sure that the list of programs is first received from MP, before the donation details directive is rendered. Since there is no logic left in the directive, I deleted the jasmine test. The protractor test is covering everything.


crossroads.net/app/app.js
```diff
@@ -171,9 +171,10 @@ var _ = require('lodash');
     .directive("emptyToNull", require('./shared/emptyToNull.directive.js'))
     .directive("stopEvent", require('./shared/stopevent.directive.js'))
     .directive("requireMultiple", require('./shared/requireMultiple.directive.js'))
-    .directive("svgIcon", require('./shared/svgIcon.directive.js'));
+    .directive("svgIcon", require('./shared/svgIcon.directive.js'))
+    .directive("naturalNumber", require('./give/directives/naturalNumber.validation.directive.js'));
```
silbermm: Why add naturalNumber directive here? Shouldn't it be in the index.js of the give module?

crossroads.net/app/routes.js
```diff
@@ -180,7 +180,14 @@
             .state("give", {
                 url: "/give",
                 controller: "GiveCtrl as give",
-                templateUrl: "give/give.html"
+                templateUrl: "give/give.html",
+                resolve:{
+                  programList:  function(getPrograms){
+                    return getPrograms.Programs.get({programType: 1}).$promise.then(function(response){
```
silbermm: does this actually work? you should only need to return `getPrograms.Programs.get({programType:1}).$promise`

PR #267, opened 2015-05-07
Feature/us1315 add copy below give

saylormic:


tatelucas:
:+1:

crossroads.net/app/give/demo/guest_giver/give.html
```diff
@@ -37,7 +39,14 @@
                         </select>
                     </div>
                     <!--/form-group-->
-                    <a ui-sref="/demo/guest-giver/login" class="btn btn-primary btn-lg col-xs-12">Give $125</a>
+                    <div class="form-group">
+                      <a ui-sref="/demo/guest-giver/login" class="btn btn-primary btn-lg col-xs-12">Give $125</a>
+                      <p class="give-info">You may also manage you
```
bzmillerboy: This isn't part of the form so it shouldn't be in a form-group or the <form> element. Place it just below the <form>.

crossroads.net/app/give/demo/guest_giver/give.html
```diff
@@ -37,7 +39,14 @@
                         </select>
                     </div>
                     <!--/form-group-->
-                    <a ui-sref="/demo/guest-giver/login" class="btn btn-primary btn-lg col-xs-12">Give $125</a>
+                    <div class="form-group">
```
bzmillerboy: Buttons should not be in a form-group, see bootstrap examples.

crossroads.net/app/give/demo/guest_giver/give.html
```diff
@@ -37,7 +39,14 @@
...
+                      <a ui-sref="/demo/guest-giver/login" class="btn btn-primary btn-lg col-xs-12">Give $125</a>
+                      <p class="give-info">You may also manage you 
+                        <a href="">recurring giving</a>, 
+                        <a href="">give to someone's trip</a>, or 
+                        <a href="">learn more</a> about ways to engage financially.</p>
+                    </div>
+                    <!-- /giving information -->
                 </div>
```
bzmillerboy: It looks like there is an parent form-group that may not be needed? If I look at bootstrap examples, I'm not sure why we'd need form-groups within form-groups. Can you fix this? I think each field should have it's own form-group.

PR #269, opened 2015-05-08
Feature/us1309 corkboard listing

bzmillerboy:


crossroads.net/app/corkboard/corkboard-listings.html
```diff
@@ -0,0 +1,315 @@
...
+        <p>
+          Posts are deleted after 30 days to keep things neat and tidy.
+        </p>
+      </div><!--/corkboard-intro-->
+
+      <div class="col-xs-12 col-sm-6 col-md-4 corkboard-browse">
+
+        <h4 class="flush-top">Browse</h4>
+
+          <a href="#" class="btn btn-gray block btn-labeled push-half-bottom">Needs Posted
```
saylormic: leaving # in href will cause problems with angular routing

crossroads.net/styles/modules/buttons.scss
```diff
@@ -3,18 +3,69 @@
...
+    border-bottom-right-radius: $border-radius-base;
 
     .icon {
-      margin-top: 2px;
+      margin-top: 6px;
+      color: #fff;
+    }
+  }
+
+  &.btn-gray {
```
saylormic: I see this was really carried over from Milacron, but we should use the Bootstrap mixin for creating new buttons. 

Example:

.btn-default {
  @include button-variant($btn-default-color, $btn-default-bg, $btn-default-border);
}

bzmillerboy:
Good catch, I've fixed both of these.

danrye:
.DS_Store should be in the .gitignore and removed from the repository

bzmillerboy:
Fixed DS_Store...well Dan and Matt B did for me

PR #270, opened 2015-05-08
US1313: Implement an optionally authorized wrapper for gateway controllers to use

IngageKrizJ:
Added a new Authorized wrapper that takes two lambda expressions - one to execute when the user is authenticated, one to execute when not authenticated.


Gateway/crds-angular/Security/MPAuth.cs
```diff
@@ -37,6 +44,33 @@ protected IHttpActionResult Authorized(Func<string,IHttpActionResult> doIt )
             }
         }
 
+        /// <summary>
+        /// Execute the labmda expression "actionWhenAuthorized" if the user is authenticated, or execute the expression
```
silbermm: lambda is spelled wrong again! :facepunch:

andycanterbury:
:+1:

PR #276, opened 2015-05-12
Feature/us1318 serve signup performance improvements

silbermm:
@IngageKrizJ @shankx50 @danrye @IngageRitterS @mattbrewer please review! :pray:


Gateway/MinistryPlatform.Translation/Services/GroupParticipantService.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        private static IDbConnection DbConnection()
+        {
+            var user = Environment.GetEnvironmentVariable("MP_API_DB_USER");
+            var password = Environment.GetEnvironmentVariable("MP_API_DB_PASSWORD");
+            var mpConnectionString =
+                ConfigurationManager.ConnectionStrings["MinistryPlatformDatabaseSource"].ToString()
+                    .Replace("%MP_DB_USER%", user)
+                    .Replace("%MP_DB_PASSWORD%", password);
```
IngageKrizJ: Could use Environment.ExpandEnvironmentVariables here to simplify (then no need for Environment.GetEnvironmentVariable calls or Replace call)

Gateway/MinistryPlatform.Translation/Services/GroupParticipantService.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            var tmp = sb.Remove(sb.Length - 1, 1);
+            query = query.Replace("@participants", tmp.ToString());
+            command.CommandText = query;
+
+            return command;
+        }
+
+        private static IDbConnection DbConnection()
```
IngageKrizJ: Might want to take a look at DbConnectionNameTypeConverter in schedule-jobs (https://github.com/crdschurch/schedule-jobs/blob/master/MigrateMPData/MigrateMPData/UnityUtil/DbConnectionNameTypeConverter.cs).  This allows you to inject a dbconnection using unity.

IngageKrizJ:
PR looks good overall, but couple comments about the DB stuff above.  Also, should we be concerned about managing a DB connection like this?  Is there any kind of connection pooling we could employ (do you think it is even necessary)?  I honestly can't remember the last time I had to open/close a DB connection in a web app, so maybe this is normal for .Net.

PR #278, opened 2015-05-13
Feature/us1324 corkboard create page

saylormic:


crossroads.net/app/corkboard/post-need.html
```diff
@@ -0,0 +1,22 @@
...
+      </div>
+    </div>
+    <div class="form-group">
+      <label for="desc" class="col-sm-2 control-label">Description</label>
+      <div class="col-sm-10">
+        <textarea class="form-control" rows="4"></textarea>
+      </div>
+    </div>
+    <div class="pull-right">
+      <button class="btn btn-primary">Cancel</button>
```
bzmillerboy: Can you make this the gray-light button? This is a secondary action, let's drive emphasis to the primary action.

crossroads.net/app/corkboard/post-need.html
```diff
@@ -0,0 +1,22 @@
...
+      <button class="btn btn-primary">Delete</button>
```
bzmillerboy: What does this button do? Is it needed? If not, you can remove it.

saylormic: What button?

saylormic:
Removed the delete buttons and changed cancel buttons to light gray

tatelucas:
Looks great overall. 

A few thoughts:
* For the create forms, do we want to use the pattern where the main button is blue and the secondary buttons are grey? Save as primary and cancel, remove as btn-gray-light
* The file inputs look a little off vertically compared to the label. Text inputs and textarea look ok because they are taller than the label, but with the file input being the same height, it looks off.
* The AM/PM button is bumped right up against the minute input. It needs some padding in between.
* http://localhost:8080/#/corkboard/detail goes to the main listing page for me, not the detail page. I see in routes that was removed and "state 'corkboard.detail' url: '/detail'" was added. Is that right? It doesn't work for me either
*

tatelucas:
Looks good now.

PR #291, opened 2015-05-14
US1308: Fixing some SQL related to updated payment processor id column

IngageKrizJ:
* Renamed files to remove spaces, and add .sql extension
* Removed "<> 0" from Order By clause in possible guest donor view (was causing MP Page view to barf)
* Updated "Stripe_Customer_ID" to "Processor_ID" in DonorByContactId view


IngageKrizJ:
Jim Kriz [11:31 PM] 
sandi and drye, I did have to update views in MP to get the giving stuff working - I went back and updated the SQL files with the same changes, if you can review.  PR #291: https://github.com/crdschurch/crds-angular/pull/291

Dan Rye [11:33 PM] 
jfkriz: looks good

PR #292, opened 2015-05-14
DE151: Update the "All with ContactId" Groups Subpage View

IngageKrizJ:
Correct column names (add underscores) and add Participant_ID column


IngageMaddoxT:
:+1:

PR #295, opened 2015-05-14
fix spacing

saylormic:


saylormic:
Doesn't need review, simple CSS change

PR #305, opened 2015-05-15
Feature/us1289 track registered status donations

shankx50:


Gateway/MinistryPlatform.Translation.Test/Services/DonorServiceTest.cs
```diff
@@ -113,9 +113,9 @@ public void CreateDonationAndDistributionRecord()
                 {"Transaction_code", charge_id}
             };
 
-            var response = _fixture.CreateDonationAndDistributionRecord(donationAmt, donorId, programId, charge_id, setupDate);
+            var response = _fixture.CreateDonationAndDistributionRecord(donationAmt, donorId, programId, charge_id, setupDate, true);
 
-            _ministryPlatformService.Verify(mocked => mocked.CreateRecord(donationPageId, expectedDonationValues, It.IsAny<string>(), true));
+            _ministryPlatformService.Verify(mocked => mocked.CreateRecord(It.IsAny<int>(), It.IsAny<Dictionary<string, object>>(), It.IsAny<string>(), true));
```
IngageKrizJ: Why removing the more specific Verify in favor of It.IsAny on these?  Probably could leave the specific values, but would need to update the "expectedDonationValues" dictionary to contain the "Registered_Donor" value.

CI/SQL/US1289-AddRegisteredDonorColumnToMPDonationsPage.sql
```diff
@@ -0,0 +1,19 @@
+USE [MinistryPlatform]
+GO
+
+UPDATE [dbo].[dp_Pages]
```
IngageKrizJ: On these scripts, Dan and Sandi decided on an approach where instead of UPDATing an existing row, we would DELETE the existing row, then INSERT it back with desired values.  I'm not sure if that could be done here, since this Page is probably referenced by foreign keys in Page views, Sub-Page views, etc, but I still thought I should mention it.

IngageKrizJ:
:+1:

PR #310, opened 2015-05-18
DE164

IngageMaddoxT:


andycanterbury:
:+1:   Code looks good. Ran it in localhost and saw opportunities other than the first, working properly.

PR #317, opened 2015-05-20
US1409: Updated to set maxlength html attribute on amount input

IngageKrizJ:
This will prevent user from even typing more digits than can be accepted


shankx50:
:+1:

PR #323, opened 2015-05-22
Feature/us1385 signup to serve error msgs

silbermm:
@danrye ,

This was the work of displaying the correct contextual messages and global message. Previously we were showing a very detailed global message and no contextual messages.

### DIFF
--- a/Gateway/crds-angular/Services/ServeService.cs
+++ b/Gateway/crds-angular/Services/ServeService.cs
@@ -221,7 +221,7 @@ namespace crds_angular.Services
                         }
                         catch (ApplicationException ex)
                         {
-                            logger.Debug("there is no need to remove the event participant because there is not one");
+                            logger.Debug(ex.Message + ": There is no need to remove the event participant because there is not one.");
                         }
                     }
                     var comments = string.Empty; //anything of value to put in comments?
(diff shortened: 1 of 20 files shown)


Gateway/crds-angular/Services/ServeService.cs
```diff
@@ -213,8 +214,15 @@ public Capacity OpportunityCapacity(int opportunityId, int eventId, int? minNeed
...
-                        //if there is already a participant, remove it because they've changed to "No"
-                        _eventService.unRegisterParticipantForEvent(participant.ParticipantId, e.EventId);
+                        try
+                        {
+                            //if there is already a participant, remove it because they've changed to "No"
+                            _eventService.unRegisterParticipantForEvent(participant.ParticipantId, e.EventId);
+                        }
+                        catch (ApplicationException ex)
+                        {
+                            logger.Debug("there is no need to remove the event participant because there is not one");
```
danrye: Add the exception to the log as well

crossroads.net/app/my_serve/serveTeam.directive.js
```diff
@@ -70,12 +80,33 @@
...
+
+      }
+
       function closePanel() {
         scope.isCollapsed = true;
       }
 
       function displayEmail(emailAddress) {
-        if (emailAddress === undefined) {
+        if (emailAddress === undefined || emailAddress === null ) {
```
danrye: Looks like you are checking for undefined, null and length > 0 below here.  Perhaps consider just using !emailAddress which would check for all those things and 0, nil, and nan.

danrye:
looks good, I'd rather us use ng-message where we are using ng-show and ng-bind, but perhaps this was done for a better UX

PR #338, opened 2015-05-28
US1109 RefactorDateFilter

silbermm:
* Now sharing last filtered date with the refine directive
* Clicking LoadMore button updates last filtered date
* Changed layout of the filter, but still needs UI/UX to refine it
* Added 'apply' button to the Dates filter and moved submit logic to it


IngageBrewerM:
:+1:

PR #340, opened 2015-05-28
Feature/us1229 existing giver without change

shankx50:


Gateway/crds-angular/Controllers/API/DonorController.cs
```diff
@@ -8,6 +8,7 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Runtime.InteropServices;
```
IngageKrizJ: What is this for?  Unused import?

jpc11623:
Looks good -- New Team

PR #341, opened 2015-05-29
Feature/us1336 refactor request group

IngageMaddoxT:


Gateway/MinistryPlatform.Translation/Services/OpportunityServiceImpl.cs
```diff
@@ -177,6 +202,23 @@ public DateTime GetLastOpportunityDate(int opportunityId, string token)
...
+                var comments = string.Format("Request on {0}", DateTime.Now.ToString(CultureInfo.CurrentCulture));
+                var values = new Dictionary<string, object>
+                {
+                    {"Response_Date", DateTime.Now},
+                    {"Opportunity_ID", opportunityResponse.OpportunityId},
+                    {"Participant_ID", participant},
+                    {"Closed", false},
+                    {"Comments", comments}
+                };
+                var returnVal = _ministryPlatformService.CreateRecord("OpportunityResponses", values, apiLogin(), true);
```
jpc11623: Function is void, but there is an unused return val

PR #348, opened 2015-05-30
DE196 Fix 'TO' field not defaulting to last opp date

silbermm:
When a volunteer chooses 'NO' there roleId is set to 0, so we need to
send a valid roleId to the server when populating dates. If the roleId
is 0, get the first valid roleId in the list of roles


danrye:
yuck...

PR #352, opened 2015-06-01
Feature/us1392 layouts

bzmillerboy:


crossroads.net/styles/layout/footer.scss
```diff
@@ -14,5 +14,5 @@ footer h5 {
 }
 
 footer .open, footer .office {
-  border-left: 1px solid #878785;
-}
\ No newline at end of file
+  //border-left: 1px solid #fff;
```
saylormic: If this is commented out, should this class just be deleted?

saylormic:
:+1:

PR #355, opened 2015-06-02
DE186

silbermm:
* WHen displaying errors, I needed to check that recordId != "-1"


crossroads.net/app/mp_tools/signup_to_serve/signupToServe.controller.js
```diff
@@ -46,6 +46,7 @@
         vm.ready = true;
       });
       populateDates();
+      //console.log(vm.params);
```
IngageMaddoxT: can this be removed?

PR #359, opened 2015-06-03
make buttons larger, switch to secondary colors for cancel and calend…

tatelucas:
…ar btns

### DIFF
--- a/crossroads.net/styles/modules/buttons.scss
+++ b/crossroads.net/styles/modules/buttons.scss
@@ -6,13 +6,16 @@
   // line-height: ensure even-numbered height of button next to large input
   @include button-size($padding-large-vertical, $padding-large-horizontal, $font-size-large, $line-height-large, $border-radius-large);
 }
+
 .btn-sm {
   // line-height: ensure proper height of button next to small input
   @include button-size($padding-small-vertical, $padding-small-horizontal, $font-size-small, $line-height-small, $border-radius-small);
 }
+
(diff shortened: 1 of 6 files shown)


crossroads.net/styles/modules/buttons.scss
```diff
@@ -83,8 +86,14 @@
 .btn-gray {
   @include button-variant($btn-default-color, $gray, $gray);
 }
+
 .btn-gray-light {
   @include button-variant($gray, #f1f1f1, $gray-lighter);
+
+  &.btn-bordered {
```
bzmillerboy: Thoughts on this method? This breaks all other button styles if used in this context, so only the btn-gray-light works as an input group.

tatelucas: I didn't want to make it a general rule that any button in an input group would have a border, because I assume there would be instances where we wouldn't want a border. But maybe I'm wrong. 

We could either do it that way or I can add the .btn-bordered to all btn versions.

bzmillerboy: I think we want a border any time a btn is attached to an input box because
all input boxes have borders. If you look at the style guide, you'll the
date picker example is still using the orange button without a border, and
it doesn't look right.

Let's make all buttons as a input-group-addon have the border, regardless
of which button color you're using. Can you also update the style guide and
all other date pickers on the site to use the gray button color instead of
orange?


-Brian

PR #387, opened 2015-06-10
Defect/de182 add hiddento event rooms

IngageRitterS:
Script added for the custom field "Hidden" on the Event_Rooms table


silbermm:
:ok_woman:

PR #390, opened 2015-06-10
setting the view to credit card if that is the default source

IngageRitterS:
Set the view to be "cc" for credit card when the default source is credit card.


IngageBrewerM:
:+1:

PR #391, opened 2015-06-10
DE191 Using ng-if to show email field with or without autofocus

shankx50:
I admit it is somewhat of a creative fix, but this was the only solution that worked for me.

### DIFF
--- a/crossroads.net/app/register/register_form.html
+++ b/crossroads.net/app/register/register_form.html
@@ -3,7 +3,7 @@
     <form id="nav_login" name="registerForm" role="form" ng-submit="register()" novalidate>
 
         <div class="soft-half-bottom" ng-class="{'has-error': firstnameError() }">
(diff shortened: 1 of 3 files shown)


crossroads.net/app/register/register_form.html
```diff
@@ -3,7 +3,7 @@
     <form id="nav_login" name="registerForm" role="form" ng-submit="register()" novalidate>
 
         <div class="soft-half-bottom" ng-class="{'has-error': firstnameError() }">
-              <input id="{{passwordPrefix}}-firstname" class="form-control" id="firstname" name="firstname" type="text" ng-model="newuser.firstname" required placeholder="First Name" autofocus/>
```
IngageBrewerM: Why was the closing / removed from this element?

IngageBrewerM:
Did you try this solution in multiple browsers?

crossroads.net/app/register/register_form.html
shankx50: I did test it in Chrome, Safari and Firefox

The closing '/' is not required for an HTML5 input tag, but it is for XHTML, so it's really a matter of personal preference.

IngageBrewerM:
:+1:

PR #400, opened 2015-06-12
us1287

IngageMaddoxT:


Gateway/crds-angular/Controllers/API/VolunteerApplicationController.cs
```diff
@@ -0,0 +1,61 @@
...
+        //            return Ok(family);
+        //        }
+        //        catch (Exception exception)
+        //        {
+        //            var apiError = new ApiErrorDto("Volunteer Application GET Family", exception);
+        //            throw new HttpResponseException(apiError.HttpResponseMessage);
+        //        }
+        //    });
+        //}
+    }
```
andycanterbury: Remove commented method.

Gateway/crds-angular/Services/PersonService.cs
```diff
@@ -47,6 +47,40 @@ public List<Skill> GetLoggedInUserSkills(int contactId, string token)
             return GetSkills(contactId, token);
         }
 
+        public Person GetPerson(int contactId)
```
andycanterbury: Why aren't we just returning the Contact object if it has all the stuff in the right format?

IngageMaddoxT: Cause we don't want to expose MP objects in Gateway

andycanterbury:
:ok_hand:

PR #411, opened 2015-06-12
Added truncate filter for US72

jpc11623:


andycanterbury:
We suggest a unit test for this functionality, unless it is really time consuming to test filters.

PR #423, opened 2015-06-16
Feature/us1218 browser sync

charlie-retzler:


danrye:
Nice work @charlie-retzler

PR #424, opened 2015-06-16
DE239

silbermm:
* Created a function to handle whether the field should show error state


jpc11623:
Reviewed, looks okay, but maybe the function could be pulled out into common code at some point?

PR #438, opened 2015-06-18
Feature/us1530 give flow account state reset

IngageKrizJ:
Reset account state after giving successfully; prevent deep-linking into the give flow when the flow is not properly initialized.


IngageBrewerM:
:+1:

PR #440, opened 2015-06-18
US1502: Modify the layout of the corkboard image to be larger per UX.

charlie-retzler:
Please review both crds-corkboard and crds-angular pull requests under Feature/US1502 view a post


IngageBrewerM:
:+1:

PR #455, opened 2015-06-22
Feature/us1510 blog category page

saylormic:


crossroads.net/app/blog/blog-index.html
```diff
@@ -0,0 +1,133 @@
+<div class="col-md-8">
```
bzmillerboy: This needs a wrapper div with class "row". https://www.dropbox.com/s/mzzf5d8222cjq7h/Screenshot%202015-06-22%2014.28.25.png?dl=0

bzmillerboy: Should be col-sm-8 to match sidebar class. SM is usually used over MD, our vertical break and menu break is on MD, not SM.

crossroads.net/styles/modules/icon.scss
```diff
@@ -8,7 +8,7 @@
   height: 15px;
 }
 
-.icon-medium {
+.icon-medium, .icon-md {
```
bzmillerboy: Do we need both of these class names? Seems redundant. I'd like more descriptive classes (ex. medium over md).

saylormic: These were pulled from milacron, but this is confusing since the bootstrap standards are to use xs, sm, md, lg, xl

crossroads.net/app/blog/blog-index.html
```diff
@@ -0,0 +1,133 @@
...
+  <ul class="posts">
```
bzmillerboy: It looks like there was previously defined CSS on this class that gives this a 90% width. Can we remove that and sort the spacing out with the Bootstrap grid? Use an offset of 1 on the sidebar if the post needs more space on the right side.

bzmillerboy: The 90% is causing this on mobile. https://www.dropbox.com/s/tcuh4igf1f4vzbl/Screenshot%202015-06-22%2014.42.35.png?dl=0

PR #463, opened 2015-06-22
Feature/us1511 single blog post

saylormic:


crossroads.net/app/blog/blog-post.html
```diff
@@ -0,0 +1,75 @@
+<div class="row">
+  <div class="col-sm-8">
+
+    <h2>Kid City Water Slides Opening Soon</h2>
+    <em>
```
bzmillerboy: I think the post meta should be wrapped in a <p> like it is on the blog category page. This way it has some bottom margin.

PR #464, opened 2015-06-22
Defect/de254 missing age in page view

IngageMaddoxT:


Gateway/crds-angular/MinistryPlatform.local test.config
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
danrye: Did you mean to check this in?

PR #465, opened 2015-06-23
Script to add email address to page view 1058

IngageRitterS:
Adding a script that was overlooked


IngageRitterS:
Jim Kriz [7:55 AM]7:55 - for DE252, there actually is already a SQL script, but it looked like it somehow had not run (US1260-UpdateOfDonorByContactIdView)
I removed it from the cr_Scripts table, so the next time we run the migration, it should run

PR #469, opened 2015-06-23
Defect/de255 bookmarked pages in give flow

IngageKrizJ:
Change give flow to use a single URL; updated protractor tests to account for this


IngageBrewerM:
:+1:

PR #470, opened 2015-06-24
Defect/de257 give page wonkiness

IngageRitterS:
On the change page, when the Amount box was in error state, the red highlighting was weird.  Adjusted the styling to match the amount page.


IngageKrizJ:
:+1:

PR #477, opened 2015-06-25
Feature/us1390 change payment type

IngageRitterS:
Note the the ability to actually change the payment type was already functional at this point.
Adding a protractor test for the payment change functionality.
Discovered a hardcoded values in confirm.html and corrected it.
Correcting the error handing of the initiative box on the change page.


IngageBrewerM:
:+1:

PR #483, opened 2015-06-26
DE265: Changed anchor to load homepage for section of the site you ar…

charlie-retzler:
…e currently on. Prevented changing screens per UX if you are not on a protected page.


jpc11623:
Reviewed, looks okay.

PR #484, opened 2015-06-26
US64: Correct issues with displaying newline characters in descriptio…

charlie-retzler:
…n field.

Please review both crds-angular and crds-corkboard PR's.


jpc11623:
Reviewed, looks good.

PR #486, opened 2015-06-26
Feature/us1493 give flow state reset

IngageRitterS:
Resetting the email address when giving as guest (rootscope.user is not defined).
Call initdefaultstate when tostate.name is give - this will initialize all fields when you start the giving process, then navigate away, then return to the giving process.
Add and update test accordingly.


IngageKrizJ:
:+1:

PR #490, opened 2015-06-26
Defect/de227 email from donor

IngageRitterS:
Getting email from donor.
Resolving issue with initdefaultstate.
Renaming logger to _logger because the warning was driving me nuts.


IngageKrizJ:
:+1:

PR #495, opened 2015-06-29
Fixing the sql scripts to correctly add the template ID to the commun…

IngageRitterS:
…ication ID on programs

Communication ID on the Program table contains the ID of the corresponding message template.  Previously the sql scripts only created the message template and added the communication ID field to the Program table - there was no script that update the Communication ID on the Program table to reflect that correct template.  These scripts resolve that issue.


jpc11623:
Reviewed, looks okay.

PR #508, opened 2015-07-01
Feature/us1300 the daily signup

tatelucas:


crossroads.net/app/thedaily/thedaily.html
```diff
@@ -0,0 +1,40 @@
+<div class="col-sm-8 col-sm-offset-2">
```
bzmillerboy: Needs a row class wrapper.

PR #510, opened 2015-07-01
Feature/us1501 & us1512 spike ng14 upgrade 

IngageBrewerM:


IngageBrewerM:
Created as a marker but don't want it merged to development until QA does regression, so closed.

PR #513, opened 2015-07-02
Feature/us1588 donation status

IngageRitterS:
Add new tables to MP database and write scripts
Add new page and subpage to MP and write scripts
Update translation layer to populate donation status to pending


CI/SQL/US1588-CreateDonationStatusMappingTable.sql
```diff
@@ -0,0 +1,47 @@
...
+GO
+
+/****** Object:  Table [dbo].[Donation_Status_Mapping]    Script Date: 7/1/2015 9:41:28 PM ******/
+SET ANSI_NULLS ON
+GO
+
+SET QUOTED_IDENTIFIER ON
+GO
+
+CREATE TABLE [dbo].[Donation_Status_Mapping](
```
IngageKrizJ: Sorry for the nitpicky comment - is this really a "Mapping" table?  Would "Donation_Statuses" be more appropriate?  That would match with other similar tables like Contact_Statuses, dp_Communication_Statuses, etc.

CI/SQL/US1588-CreateDonationStatusTable.sql
```diff
@@ -0,0 +1,22 @@
+USE [MinistryPlatform]
+GO
+
+SET ANSI_NULLS ON
+GO
+
+SET QUOTED_IDENTIFIER ON
+GO
+
+CREATE TABLE [dbo].[Donation_Status](
```
IngageKrizJ: Sorry for the nitpicky comment - would "Donation_Statuses" be more appropriate? That would match with other similar tables like Contact_Statuses, dp_Communication_Statuses, etc.

CI/SQL/US1588-CreateDonationStatusMappingTable.sql
IngageKrizJ: Sorry for the nitpicky comment - could we drop the "_Mapping" from this table, or make it "Donation_Status_History"?

IngageKrizJ:
Sandi, Matt mentioned the MP audit log - I'm thinking we can do away with the status table and just have two new columns (Status_ID and Status_Note) on the donation table.  Apparently you can see all changes in the audit log, so that should meet the need (assuming Kathryn and Finance are ok with this as well)

IngageBrewerM:
:+1:

PR #522, opened 2015-07-06
Feature/us1582 roles

danrye:


Gateway/crds-angular/Controllers/API/LoginController.cs
```diff
@@ -40,6 +33,7 @@ public IHttpActionResult isAuthenticated()
                 {
                     //var personService = new PersonService();
                     var person = _personService.GetLoggedInUserProfile(token);
+                    var roles = _personService.GetLoggedInUserRoles(token);
```
IngageKrizJ: Should this be moved to the else block below (so we don't call it at all if we don't get a person)?

IngageKrizJ:
:+1:

PR #532, opened 2015-07-06
Feature/us68 admin remove post

danrye:


crossroads.net/core/services/session_service.js
```diff
@@ -7,11 +7,12 @@
```
silbermm: * I don't see any 'getters' for the roles. 
* Are the roles put into the cookies comma separated?
* When retrieving roles from this service, I would expect an array of roles to be returned
* Assuming you do add a 'getter', you will also need a test

PR #545, opened 2015-07-07
DE299: Fixed scope of cookies for corkboard to allow logging out afte…

charlie-retzler:
…r logging in from main site


silbermm:
:boom:

PR #547, opened 2015-07-08
Defect/de291 change page and debounces

IngageRitterS:
Styling credit card errors correctly when  submitting change page with invalid data
Removing debounces from bank info
Removing debounces from credit card info
Setting focus correctly when changing credit card info on the change page
Fixing tests
Eliminating console errors


IngageKrizJ:
:+1:

PR #566, opened 2015-07-09
DE306: Fix issues with email validation - get rid of duplicate email element

IngageKrizJ:


crossroads.net/core/email_field/email_field_directive.js
```diff
@@ -36,6 +36,12 @@ require('./email_field.html');
 
                 attachUniqueEmailFunctions(scope, emailElement, ngModel, $http, Session, User);
 
+                if(scope.focused && scope.focused == true) {
+                  $timeout(function() {
```
IngageBrewerM: Shouldn't there be a timeout value specified in this function?

IngageKrizJ: No, default is 0, which is what we want:
https://docs.angularjs.org/api/ng/service/$timeout

IngageBrewerM: Gotcha. Mind reviewing exactly why you need to do this with me in the morning? :confused:

IngageKrizJ: Yep, no problemo.

IngageBrewerM:
:+1:

PR #580, opened 2015-07-16
Feature/us1638 actionable buttons

saylormic:


crossroads.net/app/give/account.html
```diff
@@ -28,7 +28,7 @@
 	    <bank-credit-card-details></bank-credit-card-details>
 	</ng-form>
    <p class="text-muted small">You are giving <strong>{{give.amount | currency}}</strong>  to <strong>{{give.program['Name']}}</strong>. <a href="" ng-click="give.processChange()">Change</a></p>
-    <button ng-click="give.submitBankInfo()" ng-class="{disabled: give.processing}" class="btn btn-primary btn-lg col-xs-12">
+    <button ng-click="give.submitBankInfo()" ng-class="{disabled: give.processing}" class="btn btn-primary btn-lg col-xs-12 col-sm-2">
```
bzmillerboy: This button should not use a grid class, it's fixing the width therefore the button does not expand with the content.

bzmillerboy:
The login button should follow the btn-block-mobile pattern as well (it's part of the give flow).

bzmillerboy:
I don't think we want the thick bottom border on button groups. I think we need to add a   border-width: 1px; to .btn-group > .btn.

crossroads.net/styles/bootstrap/mixins/_buttons.scss
```diff
@@ -7,6 +7,7 @@
   color: $color;
   background-color: $background;
   border-color: $border;
+  border-bottom: 3px solid darken($background, 15%);
```
bzmillerboy: This is the bootstrap source file, you NEVER want to edit these files. You'll need to override bootstraps styles in our own buttons.scss file.

crossroads.net/styles/bootstrap/mixins/_buttons.scss
```diff
@@ -15,8 +16,7 @@
   &.active,
   .open > &.dropdown-toggle {
     color: $color;
-    background-color: darken($background, 10%);
-        border-color: darken($border, 12%);
+    background-color: darken($background, 15%);
```
bzmillerboy: This is the bootstrap source file, you NEVER want to edit these files. You'll need to override bootstraps styles in our own buttons.scss file.

crossroads.net/styles/bootstrap/mixins/_buttons.scss
bzmillerboy: I thought we landed on 2px bottom border, not 3px? I think the 2px looks better. https://slack.com/ is a good example.

bzmillerboy:
I think we have to fix the serve name buttons and assure they do not display the bottom border. Possibly the same fix as you implement for button groups. Also, there is an issue with the selected button displaying the wrong bottom border color, https://www.dropbox.com/s/kp68gg563gmlg8f/Screenshot%202015-07-16%2014.39.47.png?dl=0

PR #583, opened 2015-07-16
US486: Script Declined Donations View in MP

IngageKrizJ:
Created the view in MP, extracted the SQL, then deleted the view and ran this script to test.


jpc11623:
Reviewed, looks okay.

PR #586, opened 2015-07-16
Feature/us1639 check help

saylormic:


crossroads.net/app/give/bankInfo.html
```diff
@@ -1,14 +1,12 @@
 <ng-form name="bankAccountForm"  ng-model="bankAccount">
      <section class="bank-section">
-        <div class="form-group row col-md-8 col-md-offset-2" collapse="!isCollapsed">
-            <svg viewBox="0 0 512 280" class="icon icon-crossroads_check give-check-icon">
-                <use xlink:href="#crossroads_check"></use>
-            </svg>
+        <div class="form-group row col-md-8 col-md-offset-2">
```
bzmillerboy: Row and col classes can't be on the same element. This needs wrapped in a row class.

bzmillerboy: I think you need a class for small screens so it's not spanned across the entire screen, something like col-md-6 col-md-offset-3 col-xs-8 col-xs-offset-2

bzmillerboy: I don't think you need form-group for this element.

PR #599, opened 2015-07-20
Feature/us1635 give button redesign

saylormic:


IngageKrizJ:
@saylormic, There is also a Give button on change.html and confirm.html, not sure if those should have the same changes applied or not.

saylormic:
Added SVG to change and confirm

PR #612, opened 2015-07-22
when a template error is received, log a message but do not throw the …

IngageRitterS:
…exception.  This allows for the proper user experience. 
If the exception is thrown, it appears to the user that the giving process failed.


IngageKrizJ:
Nitpicky, but how about logger.error instead of logger.debug? If we actually start using log4net properly, we'd want to see an error for this.

IngageRitterS:
Change made ..Thanks!

PR #615, opened 2015-07-23
add indentity insert to sql scripts

IngageRitterS:


IngageKrizJ:
Not needed - the original scripts are fine, we just need to make sure we remove the duplicates they created.

PR #618, opened 2015-07-24
Feature/us1630 process webhooks asynchronously

IngageKrizJ:


Gateway/Crossroads.AsyncJobs/Application/JobProcessorController.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Crossroads.AsyncJobs.Application
+{
+    public class JobProcessorController : ApiController
+    {
+        private readonly JobProcessor _jobProcessor;
+
+        public JobProcessorController(JobProcessor jobProcessor)
+        {
+            //_jobProcessor = jobProcessor;
```
charlie-retzler: Is this class needed? It looks like everything is hardcoded. If the class is needed clean up commented out code and get rid of unusted jobProcessor parameter

PR #631, opened 2015-07-24
Feature/us1641 handle different templates for cms

danrye:


crossroads.net/app/media/media-single.html
```diff
@@ -1,5 +1,4 @@
-<div class="container-fluid hard-sides">
-  <div class="row"><!--TODO: After layouts are implemented, add a container-fluid class here in place of row-->
```
IngageKrizJ: Was the wrong line removed here?  This comment seems to indicate we're supposed to be adding the container-fluid.

crossroads.net/app/media/media-single.html
```diff
@@ -45,11 +44,10 @@
         </div><!--/media-background-->
       </div><!--/media-curtain-->
 
-  </div><!--/row TODO: After layouts are implemented, add a container-fluid class here in place of row-->
-</div><!--/container-fluid-->
+  </div>
 
-<div class="container-fluid hard-sides">
-  <div class="row related-media"><!--TODO: After layouts are implemented, add a container class here in place of row-->
```
IngageKrizJ: Same as above - just making sure this is correct.

danrye: No, Brian and I did these changes together. The route, and the route level template now adds the fluid(ness)

crossroads.net/app/media/media-single.html
danrye: No, Brian and I did these changes together. The route, and the route level template now adds the fluid(ness)

crossroads.net/app/routes.js
```diff
@@ -419,11 +489,49 @@
...
+                if (notFoundRequest.pages.length > 0) {
+                  ContentPageService.page.renderedContent = notFoundRequest.pages[0].renderedContent;
+                  ContentPageService.page.pageType = '';
+                } else {
+                  ContentPageService.page.renderedContent = '404 Content not found';
+                  ContentPageService.page.pageType = '';
+                }
+              });
+            }
+            switch(ContentPageService.page.pageType){
```
IngageKrizJ: I assume this switch gets additional cases added as we implement additional templates, right?  Wonder if we should do it by convention instead, and just return $templateFactory.fromUrl('core/templates/' + ContentService.page.pageType + '.html')?

crossroads.net/core/core.controller.js
```diff
@@ -27,6 +28,12 @@
     // State Change Listeners //
     ////////////////////////////
     $scope.$on('$stateChangeStart', function(event, toState, toParams, fromState, fromParams) {
+      ContentPageService.toParams = toParams;
+      if(toState===fromState && ContentPageService.reload){
+        ContentPageService.reload = false;
+        $state.transitionTo(toState, toParams, { reload: true, inherit: true, notify: true });
+        event.preventDefault();
```
IngageKrizJ: Is this going to break anyone else who may be listening for $stateChangeStart?  I didn't bring this down yet, but I know in giving we also listen for this event (isolate scope), and also do an event.preventDefault in certain cases.  Just want to make sure this won't conflict.

IngageKrizJ:
Unit tests for any of this?  :)

crossroads.net/app/routes.js
danrye: Correct.  I considered the convention, but you'll note a slight case difference, our naming conventions for templates don't match naming conventions for PHP classes.  If there is a strong reason (perfomance or otherwise) we could instead break the template naming convention to match the PHP in these cases

danrye:
@IngageKrizJ I don't think we know how to unit test the routes.  And I've heard Brewer say we don't need to, but I'm willing if we know how, I know I don't personally.

crossroads.net/core/core.controller.js
danrye: I don't believe so, this listener was already here, and the only time we "break" is when the toState and the fromState are the same.

IngageKrizJ:
Looks good, I also pulled and ran some testing locally.

PR #637, opened 2015-07-29
using ngOptions to populate marital status

IngageMaddoxT:


crossroads.net/core/components/volunteer_applications/kc_adult_application/kidsClubAdultApplication.controller.js
```diff
@@ -26,6 +26,7 @@ var moment = require('moment');
     vm.format = 'MM/dd/yyyy';
     vm.gradeLevelSelected = gradeLevelSelected;
     vm.locationSelected = locationSelected;
+    vm.maritalStati = maritalStati();
```
danrye: @IngageMaddoxT nitpik... statuses (http://www.merriam-webster.com/dictionary/status)

crossroads.net/core/components/volunteer_applications/kc_adult_application/kidsClubAdultApplication.controller.js
```diff
@@ -79,6 +80,16 @@ var moment = require('moment');
       return false;
     }
 
+    function maritalStati() {
```
danrye: See above

danrye:
Why are these statuses being fetched from Ministry Platform?

danrye:
You could use ProfileReferenceData, though I'm sure we'd do this differently now that we've learned so much.  https://github.com/crdschurch/crds-angular/blob/development/crossroads.net/app/profile/personal/profile_personal_controller.js

IngageMaddoxT:
Performance call.  One less api call for something that's not likely to change.  And these forms are "not the long term plan", a.k.a Black Pulp

danrye:
:+1:

PR #638, opened 2015-07-29
changing format of search string to search for contact id as string, …

IngageMaddoxT:
…not int.  int was using wildcard search and returning multiple records.


jpc11623:
Reviewed, looks okay.

PR #646, opened 2015-07-30
Feature/us1545 error message mapping

IngageKrizJ:


crossroads.net/core/core.controller.js
```diff
@@ -80,6 +80,11 @@
       contentBlockRequest.contentBlocks.unshift(null); //Adding a null so the indexes match the DB
       //TODO Refactor to not use rootScope, now using ngTemplate w/ ngMessages but also need to pull this out into a service
       $rootScope.messages = contentBlockRequest.contentBlocks;
+      _.forEach(contentBlockRequest.contentBlocks, function(m) {
```
danrye: We (Matt S. and I) think this is how you'd want to do this, but we wrote it without testing.
```
MESSAGES = _.map(contentBlockRequest.contentBlocks, function(cb){
return {cb.title:cb.id}
})
```

danrye:
Otherwise, this looks good :+1:

crossroads.net/core/core.controller.js
IngageKrizJ: I tried several versions of that map, but I don't think it will do what we need (added a unit test so I could play around with it).  According to the [map](https://lodash.com/docs#map) docs:

> Creates an array of values by running each element in collection through iteratee  

In this case, we're not creating an array, we're adding name/value pairs (message title & message id) onto the MESSAGES hash.  I don't see any other lodash method that will do that.

PR #658, opened 2015-08-03
Feature/us1403 login redirect

IngageRitterS:
Getting token exp time from MP.  Passing it back to front end so session expires appropriately.  
Adding code to check from active session.  Updating and adding tests.  Removing console.log statements.


IngageBrewerM:
:+1: :+1: :smile:

PR #661, opened 2015-08-03
Feature/us1698 my trip page

andycanterbury:


Gateway/crds-angular/Controllers/API/TripController.cs
```diff
@@ -18,20 +18,37 @@ public TripController(ITripService tripService)
...
+            {
+                var apiError = new ApiErrorDto("Trip Search Failed", ex);
+                throw new HttpResponseException(apiError.HttpResponseMessage);
+            }
+        }
+
+        [AcceptVerbs("GET")]
+        [ResponseType(typeof (MyTripsDTO))]
+        [Route("api/trip/mytrips/{contactId}")]
+        public IHttpActionResult MyTrips(int contactId)
```
silbermm: this should be authenticated and then you won't need the contactId in the url

PR #663, opened 2015-08-04
Changing all anchor tags with target="_blank" to be href's instead of…

charlie-retzler:
… ui-sref's.


jpc11623:
Reviewed, looks okay.

PR #665, opened 2015-08-05
Defect/de357 email input

IngageRitterS:
Changed the account.html file to check for a valid session.  This will allow the login/guest giver to appear correctly as well as the email input field.  Also remove a check for session in the submitbankinfo function.  This code was causing a guest giver who had an error on the account page to be redirected to login, which was incorrect.


IngageBrewerM:
:+1:

PR #680, opened 2015-08-07
Fix incorrect ID in insert statement.

JoeKer1:


jpc11623:
Reviewed, looks okay.

PR #690, opened 2015-08-12
Feature/us1742 trip module

silbermm:


crossroads.net/app/index.html
```diff
@@ -102,6 +104,12 @@
 </body>
 <script src="https://js.stripe.com/v2/"></script>
```
charlie-retzler: With HTML5 mode now enabled we may want to add the leading slash for all assets (i.e. /assets/xxx.xx)

PR #696, opened 2015-08-13
Feature/us1802 media tags

saylormic:


silbermm:
@saylormic I clicked merge before I realized that once you update core to a new version, we will need to publish and then update this to use the new version of core

PR #709, opened 2015-08-17
Feature/us1739 trip participant tool

IngageMaddoxT:


CI/SQL/US1739-03-DevTool.sql
```diff
@@ -0,0 +1,17 @@
+USE [MinistryPlatform]
```
silbermm: I think this should go in the TestSQL folder because we only want this tool to show up in int

PR #716, opened 2015-08-18
Add a join and check on deposits table to see if exported.

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/defect/41297959327


IngageBrewerM:
:+1:

PR #767, opened 2015-08-28
Defect/de388 and 390 fix error page

charlie-retzler:


jpc11623:
Reviewed, looks okay.

PR #768, opened 2015-08-28
DE388 and DE390 - fixing back button for Message and Audio/Video media

charlie-retzler:


jpc11623:
Reviewed, looks okay.

PR #771, opened 2015-08-31
Feature/us1845 check scanner batch tool batches

IngageKrizJ:


crossroads.net/app/mp_tools/check_batch_processor/checkBatchProcessor.controller.js
```diff
@@ -8,19 +8,24 @@
...
     vm.params = MPTools.getParams();
+    vm.showClosedBatches = false;
 
     activate();
     //////////////////////
 
     function activate() {
-      CheckScannerBatches.query(function(data) {
-        vm.batches = data;
+      CheckScannerBatches.query({'onlyOpen': false}, function(data) {
```
IngageBrewerM: Why have the `onlyOpen` option if we are placing business logic below (filter on `status : 'notExported'`) to do this?

Gateway/crds-angular/Controllers/API/CheckScannerController.cs
```diff
@@ -40,11 +40,11 @@ public CheckScannerController(IConfigurationWrapper configuration, ICheckScanner
         }
 
         [Route("api/checkscanner/batches")]
-        public IHttpActionResult GetOpenBatches()
+        public IHttpActionResult GetBatches([FromUri(Name = "onlyOpen")] bool onlyOpen = true)
         {
             return (Authorized(token =>
             {
-                var batches = _checkScannerService.GetOpenBatches();
+                var batches = onlyOpen ? _checkScannerService.GetOpenBatches() : _checkScannerService.GetAllBatches();
```
IngageBrewerM: Why not pass `onlyOpen` directly into a `public GetBatches(bool onlyOpen)` method?  IMHO, it doesnt make it clearer to have separate methods when the variable is named with the proper context.

Gateway/crds-angular/Services/EzScanCheckScannerService.cs
```diff
@@ -23,16 +23,28 @@ public EzScanCheckScannerService(IDbConnection dbConnection, IDonorService donor
...
+            return (GetBatches(true));
+        }
+
+        private List<CheckScannerBatch> GetBatches(bool onlyOpen)
+        {
             List<CheckScannerBatch> batches;
             IDataReader reader = null;
             try
             {
+                var whereClause = onlyOpen ? "WHERE COALESCE(BatchStatus, 0) <> 1" : string.Empty;
```
IngageBrewerM: I believe this is the proper place to have the business logic for determining Open batches... [see note below](https://github.com/crdschurch/crds-angular/pull/771/files#r38357464)

IngageKrizJ: I went back and forth on this. I initially updated the endpoint to always
return all batches, and then filtered on the client side. Then I said,
"what if we need to just get open batches at some point in the .net side",
so I updated the endpoint to support it, and updated the frontend to make
two calls. Then I said, "well, that's silly to make two calls when we can
easily filter on the frontend". So that's where it is. I am torn here...

IngageBrewerM: Yeah, I get it.  I can see the benefits in both.  I prefer to make two calls if it keeps the BL in the back end.  Lets talk to the others and get their input.  Either way, we need to remove the logic redundancy, so if we want to keep this as part of the front end and make one call, lets get rid of the backend logic.  It's YAGNI right now.

IngageBrewerM:
:+1:

PR #772, opened 2015-09-01
adding new sp to filter congregations.  this will return all congrega…

IngageRitterS:
…tions without the All Sites option


IngageBrewerM:
I feel for @IngageRitterS today... so :+1:

PR #784, opened 2015-09-02
sql scripts for new check donor report

IngageRitterS:


CI/SQL/US1820-NewCheckDonorsRept.sql
```diff
@@ -0,0 +1,112 @@
...
+GO
+
+
+
+-- =============================================
+-- Description:	New report needed to find new donors that were added as a result of processing a check
+-- =============================================
+CREATE PROCEDURE [dbo].[report_CRDS_New_Check_Donors]
+
+	 @UserID varchar(40)
```
IngageKrizJ: UserID and PageID don't seem to be used - can we remove them, or are they needed by SSRS?

PR #786, opened 2015-09-02
Defect/de398 youtube video error

jpc11623:


crossroads.net/app/media/factories/youTubePlayer.factory.js
```diff
@@ -0,0 +1,23 @@
+(function() {
+  'use strict()';
+
+  module.exports = YouTubePlayerFactory;
+
+  function YouTubePlayerFactory($q, $window) {
```
charlie-retzler: I think you need to add code to do YouTubePlayerFactor.$inject at the top similar to the directive

PR #798, opened 2015-09-03
Fixing issue with US1687 where toggling back to Video causes YouTube …

charlie-retzler:
…to be stopped rather than paused.


jpc11623:
Reviewed, looks really complex.

PR #825, opened 2015-09-11
Feature/us1906 post internal scholorship

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/userstory/42094022706


Gateway/MinistryPlatform.Translation/Services/DonationService.cs
```diff
@@ -363,7 +365,10 @@ public List<GPExportDatum> GetGPExport(int depositId, string token)
                     CashAccount = result.ToString("Cash Account"),
                     ReceivableAccount = result.ToString("Receivable Account"),
                     DistributionAccount = result.ToString("Distribution Account"),
+                    ScholarshipExpenseAccount = result.ToString("Scholarship Expense Account"),
```
IngageKrizJ: We've tried to standardize and have MP views with columns using underscores instead of spaces.  If this GP Export view was one you created, can we update it to return all the columns with underscores, and then update this code?

PR #828, opened 2015-09-14
Make it so endpoint is Create not CreateOrUpdate

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/defect/42965672045


Gateway/crds-angular.test/Services/EzScanCheckScannerServiceTest.cs
```diff
@@ -259,7 +255,7 @@ public void TestGetContactDonorForCheck()
         }
 
         [Test]
-        public void TestUpdateForCreateOrUpdateDonor()
+        public void TestExitstingCreateDonor()
```
IngageKrizJ: Speling?  :) s/b "TestExistingCreateDonor"

PR #837, opened 2015-09-16
Defect/de425 trip donations on my trips

andycanterbury:


Gateway/crds-angular/Controllers/API/TripController.cs
```diff
@@ -170,14 +173,15 @@ public IHttpActionResult TripParticipant(string tripParticipantId)
...
-        public IHttpActionResult MyTrips(int contactId)
+        [Route("api/trip/mytrips")]
+        public IHttpActionResult MyTrips()
         {
             return Authorized(token =>
             {
                 try
                 {
-                    var trips = _tripService.GetMyTrips(contactId, token);
+                    var contactId = _authenticationService.GetContactId(token);
```
IngageMaddoxT: I'd rather see this call in GetMyTrips

PR #848, opened 2015-09-17
Feature/us1999 family giving history

IngageKrizJ:


Gateway/crds-angular/Services/DonationService.cs
```diff
@@ -191,6 +222,37 @@ public DonationYearsDTO GetDonationYearsForDonor(int donorId)
...
+                return (donorIds);
+            }
+
+            var household = _contactService.GetHouseholdById(donor.Details.HouseholdId);
+            if (household == null || household.HouseholdMembers == null || !household.HouseholdMembers.Any())
+            {
+                return (donorIds);
+            }
+
+            foreach (var member in household.HouseholdMembers)
```
dtkocher: Have this information returned with the house hold call

CI/SQL/US1999-AddStatementTypeIdAndHouseholdIdToDonorByContactIdView.sql
```diff
@@ -0,0 +1,8 @@
+USE [MinistryPlatform]
+GO
+
+UPDATE [dbo].[dp_Page_Views]
+SET [Field_List] = CONCAT([Field_List], ', Contact_ID_Table_Household_ID_Table.[Household_ID] AS [Household_ID]
```
dtkocher: Is this a good idea?  Just asking what happens in the case this is run more than once?

PR #849, opened 2015-09-17
DE437 Thankyou Page Title

silbermm:
* Changed title for thank you page


crossroads.net/app/trips/signup/tripsSignup.service.js
```diff
@@ -13,18 +13,18 @@
     };
 
     function activate() {
-      switch (signupService.campaign.formName) {
+      switch (signupService.campaign.nickname) {
         case 'GO NOLA Application':
```
andycanterbury: Should this be shortened to NOLA like the other items in this switch?

PR #850, opened 2015-09-17
DE438: Removing  ng-click="$event.stopPropagation() since it wasn't n…

charlie-retzler:
…ecessary after implementing this functionality in a directive.


jpc11623:
Reviewed, looks okay.

PR #861, opened 2015-09-18
US1966: Updated build process to rename robots.*.txt for a specific e…

charlie-retzler:
…nvironment to robots.txt since this can't be done in teamcity


jpc11623:
Reviewed, looks good.

PR #866, opened 2015-09-18
Defect/de443 error on import

IngageRitterS:


Gateway/crds-angular/Services/EzScanCheckScannerService.cs
```diff
@@ -60,7 +60,16 @@ public CheckScannerBatch CreateDonationsForBatch(CheckScannerBatch batchDetails)
                 {
                     var contactDonor = CreateDonor(check);
                     //Always use the customer ID and source ID from the Donor Account, if it exists
-                    var charge = _paymentService.ChargeCustomer(contactDonor.ProcessorId, contactDonor.Account.ProcessorAccountId, (int) (check.Amount), contactDonor.DonorId);
+                    var charge = new StripeCharge();
```
IngageBrewerM: You don't need to instantiate a `new StripeCharge()` here, just declare `StripeCharge charge;` so that it can be assigned by the `_paymentService.ChargeCustomer()` below.

PR #876, opened 2015-09-23
US2023: Added code to join SoundCloud base URL with serviceId

charlie-retzler:


jpc11623:
Reviewed, looks okay.

PR #877, opened 2015-09-23
DE452: Changed size to 10000 search results as this is the maximum su…

charlie-retzler:
…pported by amazon in one chunk. Also cleaned up JSCS warnings.


jpc11623:
Reviewed, looks okay.

PR #888, opened 2015-09-24
Support for Meta Data on Routes

danrye:


crossroads.net/app/media/media.routes.js
```diff
@@ -152,7 +154,10 @@
           Meta: function (SingleMedia, $state) {
             $state.next.data.meta = {
              title: SingleMedia.message.title,
-             description: ''
+             description: SingleMedia.message.description,
+             type: 'article',
+             card: 'summary',
+             image: SingleMedia.message.video.still
```
charlie-retzler: Does the image work for messages without video, music files, and for videos not tied to a message?

PR #928, opened 2015-10-02
DE477: Adding fix for routing homepage to prerender.io

charlie-retzler:


jpc11623:
Reviewed, looks okay.

PR #930, opened 2015-10-02
DE477: Correcting issue with not checking for both index.html and ind…

charlie-retzler:
…ex.php


jpc11623:
Reviews, looks good.

PR #952, opened 2015-10-08
US2408: Adding AttributeService and supporting functionality to get l…

charlie-retzler:
…ist of AttributeTypes. Refactored some old skills stuff to use ContactAttribute class

### DIFF
--- a/Gateway/MinistryPlatform.Models/AttributeType.cs
+++ b/Gateway/MinistryPlatform.Models/AttributeType.cs
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Reflection;
-using System.Security.Policy;
-using System.Text;
-using System.Threading.Tasks;
+﻿using System.Collections.Generic;
 
 namespace MinistryPlatform.Models
 {
--- a/Gateway/MinistryPlatform.Models/AttributeType.cs
+++ b/Gateway/MinistryPlatform.Models/AttributeType.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Security.Policy;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MinistryPlatform.Models
+{
+    public class AttributeType
+    {
+        public string Name { get; set; }
+        public List<Attribute> Attributes { get; private set;  }
+        public int AttributeTypeId { get; set; }
+
+        public AttributeType()
+        { 
+            Attributes = new List<Attribute>();            
+        }
+    }
+}
--- a/Gateway/MinistryPlatform.Models/ContactAttribute.cs
+++ b/Gateway/MinistryPlatform.Models/ContactAttribute.cs
@@ -1,8 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace MinistryPlatform.Models
 {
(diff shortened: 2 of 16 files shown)


Gateway/MinistryPlatform.Models/AttributeType.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Security.Policy;
+using System.Text;
+using System.Threading.Tasks;
```
andycanterbury: Are you using threading in this class definition? I'd use ReSharper to remove unneeded references.

Gateway/MinistryPlatform.Models/ContactAttribute.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
andycanterbury: Clean up unneeded references here too.

Gateway/MinistryPlatform.Models/ContactAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace MinistryPlatform.Models
+{
+    public class ContactAttribute
+    {
+        public int dp_RecordID { get; set; }
+        public string dp_RecordName { get; set; }
+        public int dp_Selected { get; set; }
+        public int? dp_FileID { get; set; }
```
andycanterbury: do you need dp_FileID? MP always returns this and the other dp_ fields but I don't know that we ever use many besides RecordID.

Gateway/MinistryPlatform.Translation/Services/AttributeService.cs
```diff
@@ -0,0 +1,61 @@
...
+            
+        }
+
+        public List<AttributeType> GetAttributeTypes(int? attributeTypeId)
+        {
+            var token = base.ApiLogin();
+
+            var filter = ",," + attributeTypeId;
+
+            var pageViewId = Convert.ToInt32(ConfigurationManager.AppSettings["AttributesPageView"]);
```
andycanterbury: You could also do this with configurationWrapper.GetConfigIntValue("AttributesPageView") not sure which is better. We should standardize how we get app settings...

Gateway/MinistryPlatform.Translation/Services/AttributeService.cs
```diff
@@ -0,0 +1,61 @@
...
+            : base(authenticationService, configurationWrapper)
+        {
+            
+        }
+
+        public List<AttributeType> GetAttributeTypes(int? attributeTypeId)
+        {
+            var token = base.ApiLogin();
+
+            var filter = ",," + attributeTypeId;
```
IngageKrizJ: You should enclose this in quotes, like `var filter = string.format(",,\"{0}\"", attributeTypeId);`

Gateway/MinistryPlatform.Models/ContactAttribute.cs
charlie-retzler: This is going away, but was old code renamed from Attribute to ContractAttribute

Gateway/crds-angular/Controllers/API/AttributeTypeController.cs
```diff
@@ -0,0 +1,31 @@
...
+            _attributeService = attributeService;
+        }
+
+        [ResponseType(typeof (List<AttributeType>))]
+        [Route("api/AttributeType/{attributeTypeId:int?}")]
+        public IHttpActionResult Get(int? attributeTypeId = null)
+        {
+            return Authorized(token =>
+            {                
+                var attributeTypes = _attributeService.GetAttributeTypes(attributeTypeId);
```
IngageKrizJ: Probably overkill in this scenario, but we've typically been creating a Service (and Interface) in crds_angular project, and have separate DTO objects in that layer.  That typically does a mapping between the MinistryPlatform.Models object to an object in crds_angular, and that model is sent to the client.  That way the client isn't directly tied to the MP model.

PR #969, opened 2015-10-13
Feature/us2408 refactor attributes

charlie-retzler:


Gateway/MinistryPlatform.Models/Attribute.cs
```diff
@@ -4,13 +4,13 @@ namespace MinistryPlatform.Models
...
     {
-        [JsonProperty(PropertyName = "attributeId")]
         public int AttributeId { get; set; }
-        [JsonProperty(PropertyName = "name")]
         public string Name { get; set; }
-        [JsonProperty(PropertyName = "category")]
         public string Category { get; set; }
-        [JsonProperty(PropertyName = "categoryId")]
-        public int CategoryId { get; set; }
+        public int? CategoryId { get; set; }
```
silbermm: JsonProperty needed?

Gateway/MinistryPlatform.Models/ContactAttributeType.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace MinistryPlatform.Models
+{
+    public class ContactAttributeType
+    {
+        public int AttributeTypeId { get; set; }
+        public List<ContactAttributeType> ContactAttributes{ get; set; }
```
silbermm: is this correct? It looks like a ContactAttributeType includes a list of ContactAttributeType's

Gateway/crds-angular.test/Services/ContactAttributeServiceIntegrationTest.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System;
```
silbermm: These tests where probably useful during development, but there is potential that they will fail when run in the future. 

To have useful integration tests, I think they should set up their own data instead of using data that is already in the database/platform.

Gateway/crds-angular/Controllers/API/AttributeTypeController.cs
```diff
@@ -18,7 +18,7 @@ public AttributeTypeController(IAttributeService attributeService)
             _attributeService = attributeService;
         }
 
-        [ResponseType(typeof (List<AttributeType>))]
+        [ResponseType(typeof (List<AttributeTypeDTO>))]
         [Route("api/AttributeType")]
```
silbermm: Our normal convention with endpoints is lowercase. 

This doesn't look like something you changed, but it is worth calling out.

Gateway/crds-angular/Controllers/API/AttributeTypeController.cs
```diff
@@ -30,7 +30,7 @@ public IHttpActionResult Get()
             });
         }
 
-        [ResponseType(typeof(AttributeType))]
+        [ResponseType(typeof(AttributeTypeDTO))]
         [Route("api/AttributeType/{attributeTypeId}")]
```
silbermm: lowercase for endpoints

Gateway/crds-angular/Services/ContactAttributeService.cs
```diff
@@ -0,0 +1,142 @@
...
+                        Notes = contactAttribute.Notes
+                    };
+
+                    results.Add(mpContactAttribute);
+                }
+            }
+            return results;
+        }
+
+        private string GetApiUserToken()
```
silbermm: should this be in the gateway layer? 

Why not create an overloaded method in the translation layer that doesn't take a token and when that method is called, use the ApiLogin() from the base service?

Gateway/crds-angular.test/Services/ContactAttributeServiceIntegrationTest.cs
charlie-retzler: Agree that these were useful for development, but will not age well. Possibly still useful for the near future, and cleaned up when the rest of the work is completed?

PR #990, opened 2015-10-15
fixing tests

IngageMaddoxT:


Gateway/crds-angular.test/Services/TripServiceTest.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using crds_angular.Models.Crossroads.Serve;
```
andycanterbury: Did you need to add this just to add the token to the calls?

PR #1016, opened 2015-10-21
Defect/de541 congregation on recurring gift

IngageRitterS:


Gateway/crds-angular/Services/DonorService.cs
```diff
@@ -189,6 +191,8 @@ public int CreateRecurringGift(string authorizedUserToken, RecurringGiftDto recu
                                                            response.id,
                                                            contactDonor.ProcessorId);
             var stripeSubscription = _paymentService.CreateSubscription(plan.Id, contactDonor.ProcessorId);
+            var contact = _mpContactService.GetContactById(contactDonor.ContactId);
+            var congregation = contact.Congregation_ID ?? 5;
```
IngageKrizJ: Can you put a comment here to indicate it is the "Not Site Specific" congregation?  (Or get it from a config file?)

PR #1019, opened 2015-10-21
Defect/de545 passport not persisting

silbermm:


Gateway/MinistryPlatform.Models/MyContact.cs
```diff
@@ -36,6 +36,6 @@ public class MyContact
         public string Passport_Middlename { get; set; }
         public string Passport_Country { get; set; }
         public string Passport_Number { get; set; }
-        public DateTime? Passport_Expiration { get; set; }
+        public string Passport_Expiration { get; set; }
```
andycanterbury: Why did we go from datetime to string? Don't we prefer more structured data?

Gateway/MinistryPlatform.Translation/Services/ContactService.cs
```diff
@@ -149,7 +149,13 @@ private static MyContact ParseProfileRecord(Dictionary<string, object> recordsDi
                 Mobile_Carrier = recordsDict.ToNullableInt("Mobile_Carrier_ID"),
                 Mobile_Phone = recordsDict.ToString("Mobile_Phone"),
                 Nickname = recordsDict.ToString("Nickname"),
-                Age = recordsDict.ToInt("Age")
+                Age = recordsDict.ToInt("Age"),
+                Passport_Number = recordsDict.ToString("Passport_Number"),
+                Passport_Country = recordsDict.ToString("Passport_Country"),
+                Passport_Expiration = ParseExpirationDate(recordsDict.ToNullableDate("Passport_Expiration")),
```
andycanterbury: I think you changed this to a string but you are still parsing it to a nullable date?

Gateway/MinistryPlatform.Models/MyContact.cs
silbermm: because that's what we were doing for Anniversary date. I originally added it as a DateTime, but wanted to follow the same convention in the entire Model

Gateway/MinistryPlatform.Translation/Services/ContactService.cs
silbermm: Same as the above AnniversaryDate. I used that as a example

Gateway/MinistryPlatform.Translation/Services/ContactService.cs
```diff
@@ -158,6 +164,15 @@ private static MyContact ParseProfileRecord(Dictionary<string, object> recordsDi
             return contact;
         }
 
+        private static string ParseExpirationDate(DateTime? date)
```
andycanterbury: This method is identical to ParseAnniversaryDate except in name. Can we make this a single ParseNullableDate() instead of two methods?

PR #1023, opened 2015-10-22
Feature/us2571 recurring gift payment source

IngageKrizJ:
* Update Create|Edit|DeleteRecurringGift methods to use Stripe customer ID from DonorAccount instead of Donor
* [US2571](https://rally1.rallydev.com/#/27593764268d/detail/userstory/45130492074)


Gateway/crds-angular/Services/DonorService.cs
```diff
@@ -181,38 +185,123 @@ public string DecryptValues(string value)
...
+
+                recurGiftId = _mpDonorService.CreateRecurringGiftRecord(authorizedUserToken,
+                                                                            contactDonor.DonorId,
+                                                                            donorAccountId,
+                                                                            EnumMemberSerializationUtils.ToEnumString(recurringGiftDto.PlanInterval),
+                                                                            recurringGiftDto.PlanAmount,
+                                                                            recurringGiftDto.StartDate,
+                                                                            recurringGiftDto.Program,
+                                                                            stripeSubscription.Id,
+                                                                            congregation);
```
IngageBrewerM: I'm going to pass this code review, but I think this method and a few others in DonorService have way too many parameters and it makes me sad. http://stackoverflow.com/questions/2244860/when-a-method-has-too-many-parameters

PR #1026, opened 2015-10-22
Feature/us2402 populate preferred name

jpc11623:


IngageRitterS:
The rsc file appears to be the wrong one.

IngageRitterS:
Are you renaming the TooManyHeads.rsc?  If so, shouldn't the old copy be deleted?

report_cr_di_NullPreferredName - it looks like this is the supporting SP for the DI report.  however, I thought we needed to look at the audit log to see what contacts had been changed?

Where is the RSC file for the new DI report part?

PR #1031, opened 2015-10-23
Feature/us216 setup a recurring gift as admin

dtkocher:
I have found an error where it is not using the impersonated donor

https://rally1.rallydev.com/#/27593764268d/detail/userstory/24467666822

http://mp-build.cloudapp.net/viewType.html?buildTypeId=Skunkworks_CrossroadsFeatureBuilds&branch_Skunkworks=feature%2FUS216-SetupARecurringGiftAsAdmin&tab=buildTypeStatusDiv

http://mp-build.cloudapp.net/viewType.html?buildTypeId=Skunkworks_TestDatabaseScripts&branch_Skunkworks=US216-SetupARecurringGiftAsAdmin&tab=buildTypeStatusDiv


Gateway/MinistryPlatform.Translation/Services/MinistryPlatformServiceImpl.cs
```diff
@@ -247,6 +248,15 @@ public void UpdateSubRecord(int subpageId, Dictionary<string, object> dictionary
             VoidCall(token, platformClient => platformClient.UpdateSubpageRecord(subpageId, dictionary, false));
         }
 
+        public UserInfo GetContactInfo(string token)
+        {
+            using (new OperationContextScope(_platformServiceClient.InnerChannel))
```
IngageKrizJ: Do you even need to do this stuff?  It looks like Call<>() method does all this, so this method could just be a single line:
```
return Call<UserInfo>(token, platformClient => platformClient .GetCurrentUserInfo());
```

PR #1033, opened 2015-10-23
DE557: Option 1 - recalculate start date to be in the future

IngageKrizJ:
* Please see description in [DE557](https://rally1.rallydev.com/#/27593764268d/detail/defect/45330642256) for more information.
* **NOTE:** _DO NOT MERGE_ until defect is discussed with Dave and UX (and whoever else would like to weigh in)
* Another possible solution is to rework how we manage the Plan and Subscription in Stripe, and then I believe we could fix without the side effects mentioned in the defect.  This change would not be a major coding effort, but would require a regression on editing recurring gifts, as we would be changing more components involved in that flow.  See [PR 1036](https://github.com/crdschurch/crds-angular/pull/1036), which implements this possible solution.
* **IMPORTANT: Do not merge this pull request AND [PR 1036](https://github.com/crdschurch/crds-angular/pull/1036).  These are alternate solutions to the same problem.**


dtkocher:
:+1: 

Looks good but will wait to see on discussion to merge.

IngageKrizJ:
Not going to implement, going with "Option 2" , which is reassigning subscription in Stripe when plan amount changes.

PR #1054, opened 2015-10-27
Switched to adding job step

jpc11623:


IngageRitterS:
@subsystem=N'TSQL' is in the SP twice

PR #1062, opened 2015-10-27
DE575 - Fix issue where person who can't impersonate is not seeing error message

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/defect/45627758511


IngageBrewerM:
:+1:

PR #1090, opened 2015-10-30
Feature/us2549 refactor skills

charlie-retzler:


crossroads.net/app/profile/skills/profileSkills.controller.js
```diff
@@ -1,53 +1,71 @@
...
+        skillsByCategory[attribute.category].skills.push(attribute);
       });
+
+      return convertHashValuesToArray(skillsByCategory);
     }
 
-    function skillTrashCan(skill) {
-      //toggle Selected
-      skill.Selected = !skill.Selected;
+    function convertHashValuesToArray(obj) {
```
IngageKrizJ: Isn't this doing the same thing as lodash values(obj)?  For example: `_.values(obj)`

Gateway/crds-angular/Services/ContactAttributeService.cs
```diff
@@ -178,6 +190,21 @@ private List<ContactAttribute> TranslateMultiToMPAttributes(Dictionary<int, Cont
             return results;
         }
 
+        private static ContactAttribute TranslateMultiToMPAttribute(ContactAttributeDTO contactAttribute, ContactAttributeTypeDTO contactAttributeType)
```
IngageKrizJ: Could you put this in AutoMapperConfig.cs?

charlie-retzler: This uses 2 objects that would need to be merged into a 3rd object. Per our discussion it's probably not worth changing at the moment.

PR #1097, opened 2015-10-30
Feature/us2586 add publication subscription

jpc11623:


CI/SQL/US2586-05-UpdatePublicationTable.sql
```diff
@@ -0,0 +1,85 @@
...
+ALTER TABLE [dbo].[dp_User_Publications] CHECK CONSTRAINT [FK_dp_User_Publications_dp_Publications]
+GO
+
+ALTER TABLE [dbo].[dp_Communication_Publications]  WITH CHECK ADD  CONSTRAINT [FK_dp_Communication_Publications_dp_Publications] FOREIGN KEY([Publication_ID])
+REFERENCES [dbo].[dp_Publications] ([Publication_ID])
+GO
+
+ALTER TABLE [dbo].[dp_Communication_Publications] CHECK CONSTRAINT [FK_dp_Communication_Publications_dp_Publications]
+GO
+
```
charlie-retzler: I'm not sure if we need to drop / add constraints back as long as the related tables are truncated or have the reference set to null.

Gateway/crds-angular/Unity.config
```diff
@@ -22,6 +22,7 @@
     <register type="crds_angular.Services.Interfaces.IEmailCommunication" mapTo="crds_angular.Services.EmailCommunication" />
     <register type="crds_angular.Services.Interfaces.IDonationService" mapTo="crds_angular.Services.DonationService" />
     <register type="crds_angular.Services.Interfaces.IProgramService" mapTo="crds_angular.Services.ProgramService" />
+    <register type="crds_angular.Services.Interfaces.ISubscriptionsService" mapTo="crds_angular.Services.SubscriptionsService" />
```
charlie-retzler: Does this need to be added to the other Unity.config file for async?

Gateway/crds-angular/Services/SubscriptionsService.cs
```diff
@@ -0,0 +1,38 @@
...
+                    publication.TryGetValue("ID", out pID);
+                    if ((int)pID == (int)spID)
+                    {
+                        subscription.TryGetValue("Unsubscribed", out unSub);
+                        break;
+                    }
+                }
+                publication.Add("Subscribed", !(bool)unSub);
+            }
+            return (publications);
```
charlie-retzler: You shouldn't need parenthesis for 'return (publications);' statement

crossroads.net/app/profile/profile.controller.js
```diff
@@ -22,6 +23,7 @@
     vm.locations = Locations;
     vm.locationFocus = locationFocus;
     vm.profileData = { person: Person };
+    vm.subscriptions = Profile.Subscriptions.query();
```
charlie-retzler: Should this be resolved in the route?

Gateway/crds-angular/Controllers/API/PublicationsController.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
charlie-retzler: I don't see anything using this class. Is this remnant of the original implementation?

Gateway/crds-angular/Unity.config
charlie-retzler: Tested and we don't need this in the async unity.config for now

CI/SQL/US2586-05-UpdatePublicationTable.sql
charlie-retzler: Discussed with John and TRUNCATE statements do indeed need constraints removed. Alternatively we could have used DELETEs and RESEED, but this works too.

PR #1113, opened 2015-11-04
US2693-remove name

IngageRitterS:
Note that pledgename was unused so I removed it as well.


IngageKrizJ:
Sandi, I know they mentioned possibly/probably needing the name for trip giving.  If we remove the name from all the DTOs and other .Net objects, will that make it unavailable for those cards?  Or do they use separate objects & services?

PR #1115, opened 2015-11-04
Feature/us2669 check number from ez scan

IngageKrizJ:
* [US2669](https://rally1.rallydev.com/#/27593764268d/detail/userstory/45697513311)
* There are more changes here than required for this story, for two reasons:
  1. I refactored DonorService.CreateDonationAndDistributionRecord to take a data object, rather than adding a 17th parameter to the method signature
  2. There was a defect in the EzScanCheckScannerService.CreateDonationsForBatch, it was expecting encrypted account and rtn, but that was not the case.  So I updated the DAO to return encrypted values, which it should have been doing from the beginning.


IngageRitterS:
Beautiful.  Thanks for doing the extra work that no one else wanted to take on!

PR #1116, opened 2015-11-04
Defect/de609 donation emails no workie

IngageKrizJ:
* [DE609](https://rally1.rallydev.com/#/27593764268d/detail/defect/46127132491)
* We broke this a couple sprints ago when we did the Recurring Gift emails
* Don't send null value from DonorService to CommunicationService in merge data
* Put safeguard in CommunicationService to check for nulls before using merge data value


IngageBrewerM:
:+1:

PR #1122, opened 2015-11-04
DE613: Make sure we don't exceed the MP limit for check number

IngageKrizJ:
* [DE613](https://rally1.rallydev.com/#/27593764268d/detail/defect/46154900660)
* Changed check scanner service to strip off leading zeroes and spaces from check number, and then take the rightmost 15 digits of the check number if it still exceeds 15 (which would be a check number in the 999 trillion range...)


IngageBrewerM:
:+1:

PR #1127, opened 2015-11-05
Defect/de608 cannot edit past recurring gift

IngageKrizJ:
* [DE608](https://rally1.rallydev.com/#/27593764268d/detail/defect/46097190300)
* Also updated Unity.config in AsyncJobs to fix error


IngageBrewerM:
:+1:

PR #1129, opened 2015-11-05
Feature/us2688 indicate childcare need

silbermm:


Gateway/crds-angular/Models/Crossroads/ParticipantSignup.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
```
andycanterbury: Is this using needed?

crossroads.net/app/common/childcare/childcare.service.js
```diff
@@ -0,0 +1,16 @@
+(function() {
+  'use strict';
+
+  module.exports = ChildCareService;
+
+  ChildCareService.$inject = ['$resource'];
+
+  function ChildCareService($resource) {
+    var childCareService = {
+      Need: $resource('__API_ENDPOINT__' + 'api/childcare/need')
```
andycanterbury: Does this endpoint exist? I don't see it in the changes on the .net side.

PR #1130, opened 2015-11-05
Defect/de615 home phone not validating

charlie-retzler:


crossroads.net/app/common/profile/household/profileHousehold.controller.js
```diff
@@ -34,6 +34,7 @@
     vm.displayLocation = displayLocation;
     vm.states = getStates();
     vm.validation = Validation;
+    vm.phoneFormat = vm.validation.phoneFormat();
```
silbermm: is this really necessary? why not just use `household.validation.phoneFormat()` in the view?

crossroads.net/app/common/profile/personal/profilePersonal.template.html
```diff
@@ -244,7 +244,7 @@
     </div>
 
     <div class="row">
-      <div class="form-group col-sm-6" ng-class="{'has-error':  profile.validation.showErrors(profile.pform, 'mobile-phone') && profile.requireMobilePhone }" >
+      <div class="form-group col-sm-6" >
```
silbermm: I think this will break the trips application. Mobile phone is required on the trip application and this will make it so it's not highlighted when in an error state

PR #1139, opened 2015-11-09
Feature/us2687 email confirm group childcare need

andycanterbury:


Gateway/MinistryPlatform.Translation/Services/GroupService.cs
```diff
@@ -272,5 +286,40 @@ public List<Group> GetGroupsForEvent(int eventId)
...
+            var confirmation = new Communication 
+            { 
+                EmailBody = emailTemplate.Body, 
+                EmailSubject = emailTemplate.Subject,
+                AuthorUserId = 7,
+                DomainId = 1,
+                FromContactId = 7,
+                FromEmailAddress = fromAddress,
+                MergeData = mergeData,
+                ReplyContactId = 7,
```
IngageMaddoxT: do we have these values in a config file?

andycanterbury: Not that I can find.

PR #1148, opened 2015-11-11
Feature/us237 view my giving statement

IngageRitterS:


Gateway/crds-angular/App_Start/AutoMapperConfig.cs
```diff
@@ -105,10 +106,20 @@ public static void RegisterMappings()
...
+                                ? PaymentType.SoftCredit
+                                : (System.Enum.IsDefined(typeof (PaymentType), src.paymentTypeId) ? (PaymentType) src.paymentTypeId : PaymentType.Other)),
                         PaymentProcessorId = src.transactionCode,
                         Name = ((src.softCreditDonorId != 0) ? src.donorDisplayName : null),
+                        
                     };
+
+                    if (src.donationAmt < 0 )
+                    {
+                        dest.Status = src.donationStatus == 1 ? (DonationStatus)1 : (DonationStatus)5;
```
IngageKrizJ: Instead of (DonationStatus)1 and (DonationStatus)5, can you just use DonationStatus.Pending, DonationStatus.Refunded?  Also, this change probably needs to be applied in AutoMapper config in Crossroads.AsyncJobs.

PR #1186, opened 2015-11-23
Feature/us2740 rsvp to one time event

silbermm:


Gateway/crds-angular.test/Models/Crossroads/Events/EventHelpers.cs
```diff
@@ -0,0 +1,54 @@
+﻿
+using System;
+using System.Reflection.Emit;
+using MinistryPlatform.Models;
+
+namespace crds_angular.test.Models.Crossroads.Events
+{
+    public class EventHelpers
```
jpc11623: Can this whole class be changed to static? Is it being initialized anywhere or are the static methods only being called on it?

PR #1187, opened 2015-11-23
media background

tatelucas:


bzmillerboy:
make this work for the audio section as well

PR #1194, opened 2015-11-25
Feature/us2729 communicate with current volunteers

silbermm:


CI/SQL/US2729-ViewForGroupAndEventParticipants.sql
```diff
@@ -0,0 +1,28 @@
...
+SET IDENTITY_INSERT [dbo].[dp_Page_Views] ON
+INSERT INTO [dbo].[dp_Page_Views]
+           ([Page_View_ID]
+		   ,[View_Title]
+           ,[Page_ID]
+           ,[Description]
+           ,[Field_List]
+           ,[View_Clause])
+     VALUES
+           (92266
```
IngageBrewerM: Why was a Page_View_ID chosen that is this large?

andycanterbury: Because it was next?

IngageBrewerM: Not 100% sure about this, but I thought we decided use the lower range for views we script since views created in MP admin will take the next available high view id?

CI/SQL/US2729-ViewForGroupAndEventParticipants.sql
```diff
@@ -0,0 +1,28 @@
+USE [MinistryPlatform]
+GO
+
+IF NOT EXISTS (SELECT * FROM [dbo].[dp_Page_Views] WHERE [Page_View_ID] = 92266)
```
IngageBrewerM: needs the new page view id here too

PR #1232, opened 2015-12-04
Feature/us2785 sync mailchimp to mp

jpc11623:


Gateway/MinistryPlatform.Models/BulkEmailSubscriberOpt.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace MinistryPlatform.Models
+{
+    public class BulkEmailSubscriberOpt
+    {
+        public string id { get; set; }
+        public string email_address { get; set; }
```
charlie-retzler: All of these properties should use PascalCase and if then need to be serialized to Json use attributes and JSON.NET for serialization / deserialization

Gateway/MinistryPlatform.Translation/Services/BulkEmailRepository.cs
```diff
@@ -164,5 +164,36 @@ private void AddMergeFields(BulkEmailSubscriber subscriber, Dictionary<string, o
...
+            foreach (var subscriberOpt in subscriberOpts)
+            {
+                UpdateContactPublication(token, subscriberOpt);
+            }
+        }
+
+        public void UpdateContactPublication(string token, BulkEmailSubscriberOpt subscriberOpt)
+        {
+            var searchString = string.Format(",\"{0}\",,,,,,,\"{1}\"", subscriberOpt.publicationID, subscriberOpt.email_address);
+            var contactPublications = _ministryPlatformService.GetPageViewRecords(_configurationWrapper.GetConfigIntValue("SegmentationBasePageViewId"), token, searchString);
```
charlie-retzler: We already have the "SegmentationBasePageViewId" in a readonly field in this class. Use it instead of re-coding this here

Gateway/MinistryPlatform.Translation/Services/BulkEmailRepository.cs
```diff
@@ -164,5 +164,36 @@ private void AddMergeFields(BulkEmailSubscriber subscriber, Dictionary<string, o
...
+            var contactPublication = contactPublications.SingleOrDefault();
+            var contactPublicationID = contactPublication.ToString("Contact_Publication_ID");
+
+            Dictionary<string, object> subscriberOptDict = new Dictionary<string, object>
+            {
+                {"Contact_Publication_ID", contactPublicationID},
+                {"Unsubscribed", (subscriberOpt.status == "subscribed" ? false : true)}
+            };
+
+            _ministryPlatformService.UpdateRecord(_configurationWrapper.GetConfigIntValue("Subscribers"), subscriberOptDict, token);
```
charlie-retzler: The GetConfigIntValue here should follow the pattern in this class to use readonly fields to store these id's

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -68,6 +69,9 @@ public void RunService()
                 var publications = _bulkEmailRepository.GetPublications(_token);
                 Dictionary<string, string> listResponseIds = new Dictionary<string, string>();
 
+                // run this first??
```
charlie-retzler: Remove this comment

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -305,5 +306,51 @@ public string CalculateMD5Hash(string input)
             }
             return sb.ToString();
         }
+
+        public bool PullOptsFromThirdParty()
+        {
+            var token = _token;
```
charlie-retzler: don't make a local variable for token, use the _token everywhere in this class as there is logic to refresh this after 10 minutes

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -305,5 +306,51 @@ public string CalculateMD5Hash(string input)
             }
             return sb.ToString();
         }
+
+        public bool PullOptsFromThirdParty()
```
charlie-retzler: I think I misunderstood your question earlier about running the change in opt Ins/Outs as the first thing in this process. I think this either needs to be a separate class or make this method private and call from the RunService method, and pass in the publications

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -305,5 +306,51 @@ public string CalculateMD5Hash(string input)
...
+            foreach (var publication in publications)
+            {
+                // query mailchimp to get list activity
+                var request = new RestRequest("lists/" + publication.ThirdPartyPublicationId + "/members?since_last_changed=" + publication.LastSuccessfulSync +
+                    "&fields=members.id,members.email_address,members.status&activity=status", Method.GET);
+                request.AddHeader("Content-Type", "application/json");
+
+                try
+                {
+                    var responseContent = client.Execute(request).Content;
```
charlie-retzler: Need to capture the HTTP result and log error if it's not an OK, add a comment to handle this better in the US2861: MP/MailChimp Synch Error Handling

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -305,5 +306,51 @@ public string CalculateMD5Hash(string input)
...
+                {
+                    _logger.Error(string.Format("Opt-in sync failed for publication {0} Detail: {1}", publication.PublicationId, ex));
+                }
+            }
+
+            return true;
+        }
+    }
+
+    public class SubscriberOperation
```
charlie-retzler: Remove the SubscriberOperation as this was removed in the US2781.

Gateway/crds-angular/Services/BulkEmailSyncService.cs
```diff
@@ -305,5 +306,51 @@ public string CalculateMD5Hash(string input)
...
+    {
+        public List<Subscriber> operations { get; set; }
+    }
+
+    public class Subscriber
```
charlie-retzler: Remove the Subscriber as this was removed in the US2781.

Gateway/MinistryPlatform.Models/BulkEmailSubscriberOpt.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace MinistryPlatform.Models
+{
+    public class BulkEmailSubscriberOpt
```
charlie-retzler: It looks like this class is being used for 2 different things. One is within the MP Translation layer. The other is really to send the data to MailChimp. We should have a separate one for MailChimp that is a DTO. It would live in the crds-angular/Models/MailChimp folder

Gateway/crds-angular/Services/BulkEmailSyncService.cs
jpc11623: I'm adding a comment to handle this in that story, as I'm not sure of what approach we are ultimately going to use.

Gateway/crds-angular/Services/BulkEmailSyncService.cs
jpc11623: I went with the second option, as it seemed more compact.

Gateway/MinistryPlatform.Models/BulkEmailSubscriberOpt.cs
jpc11623: Done. We need to remember that the auto mapper be need to be forced to load.

Gateway/crds-angular/Models/MailChimp/BulkEmailSubscriberOptDTO.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class BulkEmailSubscriberOptDTO
+    {
+        [JsonProperty("id")]
+        public string ThirdPartySystemID { get; set; }
+        [JsonProperty("email_address")]
+        public string EmailAddress { get; set; }
+        [JsonProperty("status")]
+        public string Status { get; set; }
+        public int PublicationID { get; set; }
```
charlie-retzler: Does PublicationID need to be send to MailChimp? Does it need to be a different case?

Gateway/crds-angular/Services/BulkEmailSyncService.cs
charlie-retzler: I still don't see the comment in the code. It should be something very similar to this

            var response = client.Execute(request);
            if (response.StatusCode != HttpStatusCode.OK)
            {
                // This will be addressed is US2861: MP/MailChimp Synch Error Handling 
                // TODO: Should these be exceptions?
                _logger.Error(string.Format("Failed sending batch for publication {0} with StatusCode = {1}", publication.PublicationId, response.StatusCode));
                return null;
            }

Gateway/crds-angular/Models/MailChimp/BulkEmailSubscriberOptDTO.cs
jpc11623: I added the attribute.

Gateway/crds-angular/Services/BulkEmailSyncService.cs
jpc11623: I misunderstood what you were talking about. I added it and refactored slightly to support checking for the response code.

PR #1285, opened 2015-12-18
Feature/us2663 reset password

jpc11623:


Gateway/MinistryPlatform.Translation/Services/UserService.cs
```diff
@@ -72,7 +80,7 @@ public int GetUserIdByUsername(string email)
 
         public int GetContactIdByUserId(int userId)
         {
-            var records = _ministryPlatformService.GetPageViewRecords(2194, ApiLogin(), (""+userId+","));//  GetRecordsDict(Convert.ToInt32(ConfigurationManager.AppSettings["Users"]), ApiLogin(), ("," + email));
+            var records = _ministryPlatformService.GetPageViewRecords(2194, ApiLogin(), (""+userId+","));
```
IngageKrizJ: I know this was just removing a comment, but I just noticed that the search string should be "\"" + userId + "\"," (making sure the number is enclosed in quotes)

jpc11623: I fixed this.

PR #1286, opened 2015-12-18
Feature/us2652 anniversary date

silbermm:


CI/SQL/DE232-InsertPageView-Contacts.sql
```diff
@@ -1,48 +1,63 @@
...
+		,Household_ID_Table_Address_ID_Table.Foreign_Country 
+		,Gender_ID_Table.Gender_ID
+		,Contacts.Last_Name 
+		,Contacts.Middle_Name
+		,Contacts.Maiden_Name 
+		,Marital_Status_ID_Table.Marital_Status_ID 
+		,Mobile_Carrier_Table.[Phone_Carrier] AS [Mobile_Carrier] 
+		,Mobile_Carrier_Table.[Phone_Carrier_ID] AS [Mobile_Carrier_ID] 
+		,Contacts.[__Age] AS [Age]
+		,Participant_Record_Table.[Participant_Start_Date]'
```
andycanterbury: How does this work if they have more than one participant record? Is that something we need to be concerned with?

silbermm: I think that is something we don't need to be concerned with because it will be handled by data validation tasks

PR #1295, opened 2015-12-21
DE813 - Convert time sent for trial period to Stripe to UTC.

dtkocher:
Fix to send UTC time to stripe instead of GMT as this is what they use in there API.


IngageKrizJ:
Dustin, I have changes for this as well, but don't have my PC at the moment.  I actually made this same change in several places- basically anywhere we deal with the trial end date, including on the frontend (in case the user is in a different time zone, use moment to convert it to UTC).  I had to update several unit tests as well.  I can commit this stuff later this morning, or if you want to make the same changes, go ahead.

dtkocher:
@IngageKrizJ I am going to close this and defer to your changes.

PR #1306, opened 2015-12-22
US2922: Update Group Participant Type

danrye:


jpc11623:
Really big, but okay..

PR #1320, opened 2015-12-23
DE822: Check for null or 0 to indicate no email template

IngageKrizJ:
* [DE822](https://rally1.rallydev.com/#/27593764268d/detail/defect/48815512405)
* Now checking for null or 0 in communication template, then fall back to default template


IngageBrewerM:
:+1:

PR #1343, opened 2016-01-05
US2926 Signup 2 Serve Reminder

silbermm:
* Added email template
* Added a view to responses to get needed reminders
* Created new method in serveservice to send su2s reminder emails

### DIFF
--- a/Gateway/crds-angular.test/Services/ServeServiceTest.cs
+++ b/Gateway/crds-angular.test/Services/ServeServiceTest.cs
@@ -9,6 +9,7 @@ using crds_angular.Services.Interfaces;
 using Crossroads.Utilities.Extensions;
 using Crossroads.Utilities.Interfaces;
 using MinistryPlatform.Models;
+using MinistryPlatform.Translation.Models;
 using MinistryPlatform.Translation.Services.Interfaces;
 using Moq;
 using NUnit.Framework;
@@ -34,7 +35,7 @@ namespace crds_angular.test.Services
         private Mock<ICommunicationService> _communicationService;
         private Mock<IConfigurationWrapper> _configurationWrapper;
         private Mock<IApiUserService> _apiUserService;
-        private Mock<IMinistryPlatformService> _ministryPlatformService;
+        private Mock<IResponseService> _responseService;
 
         private ServeService _fixture;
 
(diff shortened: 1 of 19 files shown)


Gateway/crds-angular.test/Services/ServeServiceTest.cs
```diff
@@ -34,6 +34,7 @@ public class ServeServiceTest
         private Mock<ICommunicationService> _communicationService;
         private Mock<IConfigurationWrapper> _configurationWrapper;
         private Mock<IApiUserService> _apiUserService;
+        private Mock<IMinistryPlatformService> _ministryPlatformService;
```
andycanterbury: How was this NOT mocking ministry platform service before??

Gateway/crds-angular/Models/Crossroads/Serve/ServeProfile.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Linq;
+using System.Web;
+
+namespace crds_angular.Models.Crossroads.Serve
+{
+    public class ServeProfile : AutoMapper.Profile
+    {
+        protected override void Configure()
+        {
+            AutoMapper.Mapper.CreateMap<Dictionary<string, object>, ServeReminder>()
```
andycanterbury: This mapper is going straight from the mp dictionary to the angular model? Elsewhere, we put the dictionary into a translation model and then map between the angular and translation layer models.

Gateway/crds-angular/Services/EventService.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using crds_angular.Models.Crossroads.Events;
+using crds_angular.Models.Crossroads.Serve;
```
andycanterbury: Is this using necessary? Its the only change in this class and doesn't seem to be needed?

Gateway/crds-angular/Models/Crossroads/Serve/ServeProfile.cs
andycanterbury: I think this mapping isn't used anymore right? Is this left over from the first design?

PR #1346, opened 2016-01-05
DE789 Trips not being Discontinued

silbermm:


Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -92,6 +92,7 @@
   <add key="Pledge_Campaigns" value="361" />
   <add key="Pledges" value="363" />
   <add key="PledgesByDonorId" value="92187" />
+  <add key ="PledgeStatusDiscontinued" value="3"/>
```
andycanterbury: Extra space between key and =. :-P

PR #1365, opened 2016-01-08
Hotfix/register controllers unity

silbermm:


IngageKrizJ:
Did this work? I tried to get this working in the past, but had no luck.
Webapi seems to do its own instantiation of controllers.

PR #1400, opened 2016-01-14
US2822: Don't update CRDS tool URLs in int anymore

IngageKrizJ:
* [US2822](https://rally1.rallydev.com/#/27593764268d/detail/userstory/47529488705)
* Most of the work for this story was [server setup defined here](https://docs.google.com/document/d/1phQ--Nq7d3mDrngCqY3gqn8ST0UGrZbZ3ziftb6uqfo/edit#heading=h.5o21fc7cikqc), but this PR is making sure we don't try to update URLs in int anymore (it's unnecessary now).


dtkocher:
@IngageKrizJ US1304-TestCheckBatchTool.sql should be deleted as well.

PR #1403, opened 2016-01-15
DE852: Forgot to force prerender.io to call back over HTTP instead of…

charlie-retzler:
… HTTPS


jpc11623:
Reviewed, looks fixed.

PR #1413, opened 2016-01-18
DE939 missing cms blocks

IngageMaddoxT:


andycanterbury:
Are there CMS changes for these content blocks or did they already exist and we just weren't using them? Int is the main place for CMS right? So we don't need to script anything?

PR #1415, opened 2016-01-18
Defect/de836 rsvp family

silbermm:
return null if there is no room...


danrye:
I don't like the catchall exception, but talked with Matt and pushing through.

PR #1420, opened 2016-01-19
update trips page title

tatelucas:


bzmillerboy:
@tatelucas In addition to this, don't forget to update the dropdown on the profile image. The main link to this page.

PR #1422, opened 2016-01-19
DE931: Corrected issue with message not displaying if no video associ…

charlie-retzler:
…ated


jpc11623:
Reviewed, looks good.

PR #1509, opened 2016-02-04
DE1086 payment type and account type are different, make sure they ma…

IngageRitterS:
…p correctly

Getting account type from donor record (which is checking, savings, credit card) and using it set payment type (Bank, Check, or Credit Card).  If the account type is Credit Card, all is well.  If the account type is Checking (we do not use Savings), then set the payment type to be Bank.

### DIFF
--- a/Gateway/crds-angular/Services/DonorService.cs
+++ b/Gateway/crds-angular/Services/DonorService.cs
@@ -331,7 +331,8 @@ namespace crds_angular.Services
                 }
 
                 var acctType = _mpDonorService.GetDonorAccountPymtType(recurringGift.DonorAccountId.GetValueOrDefault());
-                var paymentType = MinistryPlatform.Translation.Enum.PaymentType.GetPaymentType(acctType).name;
+                var acct = acctType != 3 ? 5 : acctType;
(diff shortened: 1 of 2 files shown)


Gateway/crds-angular/Services/DonorService.cs
```diff
@@ -331,7 +331,8 @@ private void SendRecurringGiftConfirmationEmail(string authorizedUserToken, int
                 }
 
                 var acctType = _mpDonorService.GetDonorAccountPymtType(recurringGift.DonorAccountId.GetValueOrDefault());
-                var paymentType = MinistryPlatform.Translation.Enum.PaymentType.GetPaymentType(acctType).name;
+                var acct = acctType != 3 ? 5 : acctType;
```
IngageKrizJ: Sandi, payment type 3 looks like it is "Coin" in system lookups - was that intended?

IngageRitterS: This is poorly named (IMO).  We are actually getting the account type from the donor account, where 3 = credit card.

IngageKrizJ: Ah yes, forgot about that.  Looks good.

PR #1519, opened 2016-02-05
DE1127 Cancelled event not shown on SU2S page

lknair2015:


CI/SQL/Views/vw_crds_Serving_Participants.sql
```diff
@@ -0,0 +1,35 @@
+USE [MinistryPlatform]
```
IngageMaddoxT: this isn't the agreed to naming convention for scripts

Gateway/MinistryPlatform.Models/GroupServingParticipant.cs
```diff
@@ -28,9 +28,10 @@ public class GroupServingParticipant
         public string ParticipantEmail { get; set; }
         public int ParticipantId { get; set; }
         public string ParticipantLastName { get; set; }
-        public string Room { get; set; }
+        public string Room { get; set; } 
         public long RowNumber { get; set; }
         public bool? Rsvp { get; set; }
         public bool LoggedInUser { get; set; }
+        public bool EventCancelled { get; set; }
```
IngageMaddoxT: why do we need this field

Gateway/MinistryPlatform.Translation/Services/GroupParticipantService.cs
```diff
@@ -110,6 +111,7 @@ private static IDbCommand CreateSqlCommand(IReadOnlyList<int> participants, long
                     WHERE ( {0} ) 
                     AND Event_Start_Date >= @from 
                     AND Event_Start_Date <= @to
+                    AND Cancelled != 1
```
IngageMaddoxT: can this be in the view?

PR #1525, opened 2016-02-08
Feature/us3174 daily opt in

jpc11623:


IngageKrizJ:
Looks good.  Needs to be tested with US3174.  That code was already merged, this is just a pre-emptive defect fix.

PR #1543, opened 2016-02-11
increase max message size DE1133

JoeKer1:
Hotfix to prod.


IngageKrizJ:
Tested in release, looks good.

PR #1544, opened 2016-02-11
DE1139: Send recurring gift times as noon UTC instead of midnight UTC

JoeKer1:
This allows gifts to process on the correct date selected by the Community Member, regardless of their timezone.

hotfix to prod


IngageKrizJ:
Tested in release, looks good.

PR #1565, opened 2016-02-19
Defect/de1049 assign donor not working

IngageKrizJ:
* [DE1049](https://rally1.rallydev.com/#/27593764268ud/detail/defect/50392689335)
* Make sure to use and/or create donor account properly when scanning a check, not specific to just using assign donor


IngageBrewerM:
:+1:

PR #1567, opened 2016-02-19
US3232: Don't send confirmation emails for recurring gifts or checks

IngageKrizJ:
* [US3232](https://rally1.rallydev.com/#/27593764268ud/detail/userstory/51025236941)
* Remove confirmation emails from recurring gifts or checks


IngageBrewerM:
:+1:

PR #1585, opened 2016-02-23
Feature/us3252 journey email invites

IngageRitterS:


Gateway/crds-angular/Services/GroupService.cs
```diff
@@ -273,6 +284,57 @@ public List<GroupDTO> GetGroupsByTypeForParticipant(string token, int participan
...
+
+        public int SendJourneyEmailInvite(EmailCommunicationDTO communication, string token)
+        {
+            var participant = GetParticipantRecord(token);
+            var groups = GetGroupsByTypeForParticipant(token, participant.ParticipantId, 19);
+            var membership = groups.Where(group => group.GroupId == communication.groupId).ToList();
+            if (membership.Count <= 0)
+            {
+                //notfound, don't send invite
+                return 1;
```
charlie-retzler: Would it be better to throw an exception here instead of returning an int?

PR #1586, opened 2016-02-23
DE1180: Making MeetingID optional

charlie-retzler:


charlie-retzler:
Sara code reviewed and approved

PR #1707, opened 2016-03-01
DE1204 group type update

nogorilla:
Also updates seats taken to default to 0


nogorilla:
corresponding PR in `core-js` https://github.com/crdschurch/crds-corejs/pull/246

PR #1754, opened 2016-03-04
Add missing BaseURL and BaseMPUrl parameters to Web.config and transforms

IngageKrizJ:
The BaseUrl parameter has been in Web.config for a while, but was never used by anything in AsyncJobs.  It was recently referenced in Gateway/crds-angular/Services/EmailCommunication.cs, which is used by AsyncJobs, so AsyncJobs was failing when it tried to get this value.


jpc11623:
Reviewed, looks good.

PR #1760, opened 2016-03-07
DE1251 - removed meeting datetime from private groups

nogorilla:


charlie-retzler:
Merged through different PR's

PR #1761, opened 2016-03-07
DE1171-Pledges table should be INNER JOIN, not a LEFT JOIN

plachmann:


jpc11623:
Reviewed, looks good.

PR #1791, opened 2016-03-09
US3362 Volunteer Equipment

silbermm:


CI/SQL/2016030910600_US3362-equipment-attribute-type.sql
```diff
@@ -0,0 +1,16 @@
+USE [MinistryPlatform]
```
IngageMaddoxT: I think this file name is missing a character.  Worried about scripts running out of order.

PR #1804, opened 2016-03-10
Fix to show Distribution Account of distribution and not the gl account

dtkocher:
https://rally1.rallydev.com/#/22244455064d/detail/defect/52500254310


IngageKrizJ:
@dtkocher, I reviewed, and I don't understand the changes (could be I just don't understand what data is supposed to come from where...)  Can we talk when you get in?

IngageKrizJ:
:+1:

PR #1823, opened 2016-03-10
DE1276 adjust matching logic

IngageRitterS:


Gateway/crds-angular/Services/GroupSearchService.cs
```diff
@@ -154,8 +154,9 @@ private Boolean MatchGroupType(ObjectSingleAttributeDTO gender, ObjectSingleAttr
             //TODO I can't remember if we wanted to force married couples with other married couples or not
```
charlie-retzler: Can you remove this leftover TODO?

PR #1831, opened 2016-03-11
DE1280 change property name so decrement works

IngageRitterS:


danrye:
Not merging, @nogorilla changed his logic

PR #1846, opened 2016-03-14
DE1306 Update Group Connector Page Defintion

IngageMaddoxT:
This is for Matt S


silbermm:
Your build is failing. I think this is because this table got into demo and when we copy down demo, the create table fails. We probably made a change to the script, so we should rewrite the change as an Alter

PR #1847, opened 2016-03-14
Group finder defects

nogorilla:
Covers: 
* DE1275
* DE1285
* DE1287
* DE1298
* DE1304 (comment lists 1340 but that is incorrect)


crossroads.net/app/group_finder/services/results.service.js
```diff
@@ -52,6 +53,10 @@
                 });
               }
             });
+
+            if (_.has(group.singleAttributes, '75' ) ) {
```
charlie-retzler: I think you also need to check the value of 7017

crossroads.net/app/group_finder/services/results.service.js
```diff
@@ -52,6 +53,10 @@
                 });
               }
             });
+
```
charlie-retzler: Just above this is the issue where we are using substr instead of  indexOf !== -1. I suggest we change the type.name.substr('dog') to (type.name.indexOf('dog') !== -1.

charlie-retzler: Also need to remove the if (attribute.name === 'Kids') codeblock above since that would be for a multi-attribute and this is a single attribute.

PR #1857, opened 2016-03-15
Feature/us3346 indicate org association

silbermm:
[US3346](https://rally1.rallydev.com/#/22244455064d/detail/userstory/51924329177)

I noticed that by default ui-bootstrap is case sensitive, so going to
http://localhost:3000/go-volunteer/cincinnati/Crossroads/profile was
hitting a different route than
http://localhost:3000/go-volunteer/cincinnati/crossroads/profile

ui-router has a way to fix this by using:
`$urlMatcherFactory.caseInsensitive(true);`

* Added OrganizationService and Tests in the Translation and Gateway
layers
* Added MinistryPlatform and Crossroads Organization Models
* Added GoVolunteer Controller and Unity Config
* Added the Organizations Service to make calls to the organization
endpoint
* Added Organization promise to the routes for organizations
* Added tests for the GoVolunteerController
* Changed what the GetOrgByName function returns when null org (NotFound)


Gateway/crds-angular/Controllers/API/GoVolunteerController.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var org = _organizationService.GetOrganizationByName(name);
+            if (org == null)
+            {
+                return NotFound();
+            }
+            return Ok(org);
+        }
+    }
+}
```
IngageMaddoxT: like to see some error handling in the controller.  what if service calls throws error?

Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -101,6 +101,7 @@
   <add key="OneTimeEventRsvpTemplate" value="13714"/>
   <add key="OpportunityPage" value="348"/>
   <add key="OpportunityResponses" value ="382"/>
+  <add key="OrganizationsPage" value="10"/>
```
IngageMaddoxT: add this to other MinistryPlatform.config file.  they still aren't synced.

Gateway/MinistryPlatform.Translation/Services/OrganizationService.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public MPOrganization GetOrganization(string name, string token)
+        {
+            var search = string.Format("{0},", name);
+            var result = _ministryPlatformService.GetRecordsDict(_configurationWrapper.GetConfigIntValue("OrganizationsPage"), token, search);
+            if (result.Count > 1)
+            {
+                throw new MultipleRecordsException();
+            }
+            return MapOrganizations(result).FirstOrDefault();
```
IngageMaddoxT: .SingleOrDefault() would achieve similar results without the need for the xtra IF statement

PR #1880, opened 2016-03-17
Feature/us3187 stripe transfer bank account errors

IngageKrizJ:
* [US3187](https://rally1.rallydev.com/#/22244455064d/detail/userstory/50702717652)


Gateway/crds-angular/Services/StripeEventService.cs
```diff
@@ -128,19 +129,31 @@ public TransferPaidResponseDTO TransferPaid(DateTime? eventTimestamp, StripeTran
...
+
+            foreach (var charge in sortedCharges)
             {
                 try
                 {
-                    var  paymentId = "";
-                    if (charge.Type == "refund")
+                    var paymentId = charge.Id;
+                    StripeRefund refund = null;
+                    if ("refund".Equals(charge.Type)) // Credit Card Refund
```
IngageBrewerM: consider charge.Type.Equals("refund")

Gateway/crds-angular/Services/StripeEventService.cs
```diff
@@ -128,19 +129,31 @@ public TransferPaidResponseDTO TransferPaid(DateTime? eventTimestamp, StripeTran
...
                     {
-                        var refund = _paymentService.GetChargeRefund(charge.Id);
+                        refund = _paymentService.GetChargeRefund(charge.Id);
                         paymentId = refund.Data[0].Id;
                     }
-                    else
+                    else if ("payment_refund".Equals(charge.Type)) // Bank Account Refund
                     {
-                        paymentId = charge.Id;
+                        var refundData = _paymentService.GetRefund(charge.Id);
```
IngageBrewerM: consider `charge.Type.Equals("payment_refund")`

Gateway/crds-angular/Services/StripeEventService.cs
```diff
@@ -261,6 +256,57 @@ public TransferPaidResponseDTO TransferPaid(DateTime? eventTimestamp, StripeTran
             return (response);
         }
 
+        private DonationDTO HandleDonationNotFoundException(StripeTransfer transfer, StripeRefund refund, string paymentId, DonationNotFoundException e, StripeCharge charge)
```
IngageBrewerM: Do we really need to break this out into a separate method?

Gateway/crds-angular/Services/StripeEventService.cs
IngageKrizJ: And if charge.Type is null, this will result in a NullPointerException - this way does not  :)

Gateway/crds-angular/Services/StripeEventService.cs
IngageKrizJ: Ditto from above

Gateway/crds-angular/Services/StripeEventService.cs
IngageKrizJ: Didn't need to, but there was a bunch of code that was making the parent method very long and more complex, so this made it easier to read and less complex.

IngageBrewerM:
:+1:

Gateway/crds-angular/Services/StripeEventService.cs
IngageBrewerM: :+1:

Gateway/crds-angular/Services/StripeEventService.cs
IngageBrewerM: :+

PR #1882, opened 2016-03-18
Add uniqueness to Document Number

dtkocher:
Fix a issue found where the Document Number was not unique.


Gateway/MinistryPlatform.Translation/Services/DonationService.cs
```diff
@@ -377,14 +377,23 @@ public List<GPExportDatum> GetGPExport(int depositId, string token)
...
+                id++;
             }
 
             return gpExport;
-        } 
+        }
+
+        private GPExportDatum AddDocumentNumber(GPExportDatum datum, int index)
+        {
+            datum.DocumentNumber = String.Format("{0}000{1}", datum.DepositId, index);
```
IngageKrizJ: ReSharper will complain about this line, should use string.Format (c# alias) instead of String.Format (system type).

PR #1910, opened 2016-03-29
Feature/us3342 go cincinnati equipment

IngageMaddoxT:


crossroads.net/app/go_volunteer/page/equipment/goVolunteerEquipment.template.html
```diff
@@ -5,41 +5,31 @@
...
+               ng-value="item.attributeId"
+               id="{{item.attributeId}}"
+               ng-model="item.checked">
+        {{item.name}}
       </label>
     </div>
     <div class="checkbox">
-      <label for="equipment-other" class="control-label">
-        <input type="checkbox" value="" ng-click="otherField = !otherField" id="equipment-other">
+      <label for="7043" class="control-label">
```
andycanterbury: 7043? Is this a magic number?

IngageMaddoxT: It is a magic number.  We talked this thru the other day.  No easy way to do the "other" option.  This is the most logical solution.  The number probably doesn't matter, will be handled in c# code when save happens

PR #1939, opened 2016-04-01
Feature/brave at home

tatelucas:


dtkocher:
@tatelucas In slack Dan had the following comment: 
Dan Rye [3:29 PM] 
@tlucas: I’d really like those brave routes to be in their own file… did you try that?

Once that is answered I will merge.

PR #1949, opened 2016-04-06
Grouping change for giving by site

plachmann:
grouping on vertical axis by month and year


charlie-retzler:
Making the release and development branches off the same branch so they get the same MD5 hash

PR #1988, opened 2016-04-08
DE1395 default to individual account type for give as guest

IngageRitterS:


crossroads.net/app/common/giving/bankInfo/bankInfo.directive.js
```diff
@@ -64,6 +64,7 @@
 
           }
         }
+        scope.bankAccount.accountHolderType = 'individual';
```
charlie-retzler: Should this be in a else block? Think it may overwrite the value set above if it's a logged in user clicking the change button on a company/business account.

PR #2028, opened 2016-04-12
Fixed tests to provide better asserts

silbermm:


Gateway/crds-angular.test/Services/GoVolunteerServiceTest.cs
```diff
@@ -208,7 +208,9 @@ public void ShouldSetupMergeDataGroupConnector()
                 {"Nickname", registration.Self.FirstName},
                 {"Lastname", registration.Self.LastName}
             };
-            Assert.AreEqual(dict, mergeData);   
+            Assert.AreEqual(dict["HTML_TABLE"], mergeData["HTML_TABLE"]);
```
andycanterbury: Does this one need an assert on the count too?

PR #2036, opened 2016-04-12
DE1413 Skillz in Email -- RELEASE

silbermm:
Now the skills are included in the email if they are set


Gateway/crds-angular/Services/GoVolunteerService.cs
```diff
@@ -164,6 +164,11 @@ public bool SendMail(Registration registration)
                 listOfP.Add(BuildParagraph("Project Preference 2: ", registration.ProjectPreferences[1].Name));
                 listOfP.Add(BuildParagraph("Project Preference 3: ", registration.ProjectPreferences[2].Name));
             }
+            if (registration.Skills != null && registration.Skills.Where(sk => sk.Checked).ToList().Count > 0)
```
andycanterbury: registration.Skills.Where(sk => sk.Checked) is used here and below. Would it make sense to pull it out into a variable?

PR #2093, opened 2016-04-30
DE1457- handle if email or phone number is null

plachmann:
DE1457- handle if email or phone number is null


jpc11623:
Reviewed, looks good.

PR #2113, opened 2016-05-05
Defect/de1465 unable to copy templates

jpc11623:


Gateway/MinistryPlatform.Translation.Test/Helpers/Conversions.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MinistryPlatform.Translation.Test.Helpers
+{
+    public static class Conversions
+    {
+        public static List<int> BuildIntArrayFromKeyValue(List<Dictionary<string, object>> dictionaryList, string key)
```
IngageKrizJ: Not a big deal, since this is in a test, but could be simplified as `return dictionaryList.Select(item => item.ToInt(key)).ToList();`

PR #2118, opened 2016-05-06
Feature/us3711 add modify checkin dashboard

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/userstory/54448624475


Gateway/crds-angular/Controllers/API/EventToolController.cs
```diff
@@ -122,5 +122,37 @@ public IHttpActionResult Put([FromBody] EventToolDto eventReservation, int event
...
+                    catch (Exception e)
+                    {
+                        var msg = "EventToolController: UpdateEventRoom " + room.Name;
+                        logger.Error(msg, e);
+                        var apiError = new ApiErrorDto(msg, e);
+                        throw new HttpResponseException(apiError.HttpResponseMessage);
+                    }
+                });
+            }
+            var errors = ModelState.Values.SelectMany(val => val.Errors).Aggregate("", (current, err) => current + err.Exception.Message);
```
IngageKrizJ: I've wondered about this - how does it determine what is valid/invalid?  Not wanting to change it, just curious how it works.

crossroads.net/app/mp_tools/admin_checkin_dashboard/adminCheckinDashboard.service.js
```diff
@@ -7,7 +7,7 @@
   function AdminCheckinDashboardService($resource) {
 
     return {
-      checkinDashboard: $resource(__API_ENDPOINT__ + 'api/eventTool/:eventId/rooms'),
+      checkinDashboard: $resource(__API_ENDPOINT__ + 'api/eventTool/:eventId/rooms', null, { update: { method:'PUT' } }),
```
IngageKrizJ: If you specify one method (update) here, do you have to specify all?  Or do the others (like get) still default?  Assuming this works, just didn't know you could do it this way.

crossroads.net/app/mp_tools/admin_checkin_dashboard/eventRooms.directive.js
```diff
@@ -19,13 +20,23 @@
...
-      function update() {
+      function update(indx) {
+        return AdminCheckinDashboardService.checkinDashboard.update({eventId: scope.eventId}, scope.rooms[indx]).$promise.then(function(result) {
+                $rootScope.$emit('notify', $rootScope.MESSAGES.eventUpdateSuccess);
+                scope.rooms[indx] = result;
+              },
+
+              function(err) {
+                $log.error(err);
+                $rootScope.$emit('notify', $rootScope.MESSAGES.eventToolProblemSaving);
```
IngageKrizJ: Did you create this message (eventToolProblemSaving)?  If so, I wonder if we should create an eventToolUpdateSuccess instead of re-using the eventUpdateSuccess above.

PR #2124, opened 2016-05-10
Defect/de1473 fix validation message to not show if nothing changed

dtkocher:
This actually covers both DE1473 and DE1474.


jpc11623:
Reviewed, looks good.

PR #2163, opened 2016-05-20
US3931: Add facilitator training options

charlie-retzler:


sseissiger:
Just to be critical... the text does match the AC, but if you look at the google form, https://docs.google.com/forms/d/1o-LIE6iRTiHOFJ0bvE70srt5hlR8Nh4Lnc4PqMqdRlU/edit?usp=drive_web, it has ALL CAPS for the site names and "NEW" and "EXPERIENCED" and the times as 9:00 instead of just 9, etc. Not sure our overall standard on displaying times (which communications perfers?)

PR #2167, opened 2016-05-24
modify view_clause for pageview to show only active groups

plachmann:
modify view_clause for pageview to show only active groups


CI/SQL/20160524151100_DE1505-childcare-groups-pageview.sql
```diff
@@ -0,0 +1,8 @@
+USE [MinistryPlatform]
+GO
+
+UPDATE [dbo].[dp_Page_Views]
+          SET View_Clause = 'Groups.[End_Date] IS NULL OR Groups.[End_Date] > GetDate()'
```
silbermm: should look for end dates > a week from today

PR #2204, opened 2016-06-02
US4058 hide form if registration is closed (no available groups)

sseissiger:


crossroads.net/app/formBuilder/formBuilder.controller.js
```diff
@@ -141,6 +142,10 @@
       return (page && page.fields && page.fields.length > 1);
     }
 
+    function registrationOpen() {
```
charlie-retzler: Try to make this more generic. Possibly check FormBuilderFieldsService.hasGroupParticipant() before checking available groups. Maybe something like this

charlie-retzler: if (!FormBuilderFieldsService.hasGroupParticipant()) {
	if (vm.data.availableGroups.length < 0) {
	  return false;
	}
      }

      return true;

crossroads.net/app/formBuilder/templates/formBuilder.html
```diff
@@ -1,16 +1,22 @@
-<form ng-submit="formBuilder.save()" ng-if="formBuilder.hasForm()">
-  <div ng-repeat="field in formBuilder.page.fields">
-    <div class="row">
-      <form-field field="field" data="formBuilder.data"></form-field>
+<div ng-if='formBuilder.registrationOpen()'>
```
charlie-retzler: Rename function to make more generic

crossroads.net/app/formBuilder/templates/formBuilder.html
```diff
@@ -1,16 +1,22 @@
...
           loading-class='disabled'
           loading='formBuilder.saving' >
-  </loading-button>
+    </loading-button>
 
-</form>
+  </form>
+</div>
+
+<div ng-show='!formBuilder.registrationOpen()'>
```
charlie-retzler: same here

PR #2230, opened 2016-06-08
Changed BDay Year Masking in stored procedure

mikerobertsthrive:
Changed masking on Year from question marks to dashes in stored
procedure


jpc11623:
Reviewed, looks good.

PR #2249, opened 2016-06-09
DE1558 Don't show gaps if Once

andycanterbury:


silbermm:
this will break the way 'Once' currently works because the `generateDates` method will not run.

PR #2261, opened 2016-06-16
Feature/us4151 merge core into angular repo

charlie-retzler:


crossroads.net/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2015, Crossroads Church
```
andycanterbury: Should we update this for 2016? :stuck_out_tongue_winking_eye:

crossroads.net/spec/profile/giving/commitment/commitment_list.directive.spec.js
```diff
@@ -1,8 +1,3 @@
-require('crds-core');
-require('../../../../app/ang');
-
-require('../../../../app/app');
```
andycanterbury: Maybe still needed?

PR #2301, opened 2016-07-07
US4263: Group Participants page HTML mock

IngageKrizJ:
* [US4263](https://rally1.rallydev.com/#/27593764268d/detail/userstory/58429124813)
* Angular mocks for participants page


crossroads.net/app/group_tool/group_detail/participants/groupDetail.participants.controller.js
```diff
@@ -1,5 +1,45 @@
 export default class GroupDetailParticipantsController {
   /*@ngInject*/
-  constructor() {
+  constructor(Group, ImageService, $state) {
```
dtkocher: I might not have caught this in the last code review but I think we should name Group - GroupService.  This is more because now that we are using ES6 and going to TS we will prob. start to represent things more as Objects defined by a class.

This is more a picky thing and I don't nec. care if we do it for the mock.

IngageKrizJ: I can make that update - I think I did the same with Participant(Service).

PR #2318, opened 2016-07-11
DE1601-remove MaxParticipants column-use targetsize

plachmann:


crossroads.net/app/mp_tools/add_event_tool/addEvent.service.js
```diff
@@ -96,10 +96,10 @@
         ministryid: 2,
         congregationid: groupData.congregation.Congregation_ID,
         contactid: groupData.primaryContact.contactId,
-        startdate: moment(groupData.startDateTime).utc().format(),
+        startdate: moment(dateTime(groupData.startDate,groupData.startTime)).utc().format(),
```
silbermm: why did the startdate value change?

plachmann: While testing I noticed that the groups created were always created with the current date/time. Fixed to use date passed in with event.

PR #2321, opened 2016-07-11
Look at all groups, not just community groups

silbermm:
DE1597

* create new repo method that gets all groups
* Update tests
*

### DIFF
--- a/Gateway/crds-angular/Services/ChildcareService.cs
+++ b/Gateway/crds-angular/Services/ChildcareService.cs
@@ -251,25 +251,23 @@ namespace crds_angular.Services
             }
         }
 
+        //TODO: SPLIT OUT INTO SMALLER METHODS AND WRITE TESTS!!!!!!!!!!!!!
         public ChildcareDashboardDto GetChildcareDashboard(int contactId)
         {
             var dashboard = new ChildcareDashboardDto();
             var token = _apiUserService.GetToken();
+            
+            var contact = _contactService.GetContactById(contactId);
 
             //Figure out who is a head in my household
-            var contact = _contactService.GetContactById(contactId);
-            var household = _contactService.GetHouseholdFamilyMembers(contact.Household_ID);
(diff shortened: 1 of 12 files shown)


Gateway/crds-angular/Services/ChildcareService.cs
```diff
@@ -320,10 +318,31 @@ public ChildcareDashboardDto GetChildcareDashboard(int contactId)
...
+        /// <summary>
+        /// Determine who is the head of household in a household.
+        /// </summary>
+        /// <param name="contactId"></param>
+        /// <param name="householdId"></param>
+        /// <returns> 
+        ///     A 2-tuple with the first element being a list of all family members and the second element being a list of heads of household.
+        ///     Throws a NotHeadOfHouseholdException if the contactId passed in is not a head of household
+        /// </returns>
+        public Tuple<List<MpHouseholdMember>, List<MpHouseholdMember> >  GetHeadsOfHousehold(int contactId, int householdId)
```
plachmann: I think this needs to be split up

plachmann: I think the use of the Tuple ends up affecting readability

Gateway/crds-angular/Services/ChildcareService.cs
```diff
@@ -290,7 +288,7 @@ public ChildcareDashboardDto GetChildcareDashboard(int contactId)
                         var eventGroup = _eventService.GetEventGroupsForEvent(eventDetails.EventId, token).FirstOrDefault(g => g.GroupTypeId == _childcareGroupType);
                         var ccEventGroup = _groupService.GetGroupDetails(eventGroup.GroupId);
                         var eligibleChildren = new List<ChildcareRsvp>();
-                        foreach (var member in household)
+                        foreach (var member in members.Item1)
```
plachmann: Use of the Tuple and the .Item1 and .Item2 makes readability difficult.

PR #2340, opened 2016-07-14
Feature/us4393 unauthenticated lookup api

ilpadre:


Gateway/MinistryPlatform.Translation.Test/Services/LookupServiceTest.cs
```diff
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Configuration;
```
JoeKer1: Some of these Using are not used.

PR #2344, opened 2016-07-14
DE1619-SQL script to add columns to page view

plachmann:


CI/SQL/20160714_110000_DE1619-childcare_group_view_change.sql
```diff
@@ -0,0 +1,9 @@
+USE MinistryPlatform
+GO
+
+IF EXISTS (SELECT * FROM dbo.dp_Page_Views WHERE Page_Id=316 and Page_View_ID=2307)
```
andycanterbury: I like that you check the page view AND the page so we are less likely to clobber another team's work. :100:

PR #2347, opened 2016-07-14
Feature/US4291 bootstrap for angular2

danrye:


crossroads.net/app/boot.ts
```diff
@@ -0,0 +1,9 @@
+
+require('es6-shim/es6-shim');
+import 'reflect-metadata';
+require('zone.js/dist/zone');
```
silbermm: do we really need to `require('zone.js/dist/zone')`? There is a [discussion](https://github.com/angular/zone.js/issues/227) in the angular repo suggesting to just make it an async script tag since requiring it doesn't buy us anything...

crossroads.net/app/downgrades.ts
```diff
@@ -0,0 +1,11 @@
+import {upgradeAdapter} from './upgrade-adapter';
```
silbermm: istead of one downgrades.ts file, should we create a file per feature that we can then import into the boot.ts file? Otherwise this file will get very large.

crossroads.net/app/streaming/schedule.component.ts
```diff
@@ -0,0 +1,58 @@
+import { Component, OnInit } from '@angular/core';
+import { HTTP_PROVIDERS } from '@angular/http';
```
dtkocher: We could put this in the bootstrapping as I don't see us using another provider here.  Generally you will put things like this or ROUTER_PROVIDERS.

crossroads.net/app/streaming/schedule.component.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { Event } from './event';
+import { StreamspotService } from './streamspot.service';
+
+declare var moment: any;
+
+// TODO - placeholder for schedule if StreamspotService fails
+@Component({
+  selector: 'schedule',
+  template: `
```
dtkocher: Just for consistency I see below we are using a templateUrl?  I personally like this approach but there are advantages to making .html files for the template when it comes to making and formatting html.

crossroads.net/app/streaming/schedule.component.ts
```diff
@@ -0,0 +1,58 @@
...
+})
+
+export class ScheduleComponent implements OnInit {
+  events: Object[] = [];
+
+  constructor(private streamspotService: StreamspotService) { }
+
+  ngOnInit() {
+    this.streamspotService.byDate()
+      .then(events => {
```
dtkocher: Are we planning on using Observables or just promises right now?  Observables are more powerful, but don't know if this decision was made due to downgrading to Angular 1?

crossroads.net/app/downgrades.ts
richardbarker: I think maybe Dan can comment on this, but I'd opt for a single file.  It's only 2 lines per component.  If there are really that many, won't lots of files be even more unwieldy?

crossroads.net/app/streaming/streamspot.service.ts
```diff
@@ -0,0 +1,57 @@
...
+declare var _: any;
+
+@Injectable()
+export class StreamspotService {
+
+  //
+  // #TODO - move to ENV file?
+  //
+  private url    = 'https://api.streamspot.com/broadcaster';  // URL to web api
+  private apiKey = '82437b4d-4e38-42e2-83b6-148fcfaf36fb';
```
silbermm: we need to store the keys in environment variables and reference them here. Right now these keys have been exposed to the world and probably need to be changed.

silbermm:
I have real concerns that all of this is built without unit tests!

crossroads.net/app/boot.ts
samueljmello: I'm not opposed to this. I think this is just in there because the example had it in.

crossroads.net/app/downgrades.ts
samueljmello: For now I'd rather just have one file. That's just me personally.

crossroads.net/app/streaming/schedule.component.ts
samueljmello: Ok. I'm good to move this to the bootstrapping. Rick, Mike?

crossroads.net/app/streaming/schedule.component.ts
samueljmello: @dtkocher what do you mean? What disadvantages does temlateUrl have? I'm not sure I understand.

danrye: @samueljmello I agree.  I think the Schedule template should be pulled out into a separate file instead of inlined here.  It's easier to maintain the html files if they are in their own files.

crossroads.net/app/streaming/schedule.component.ts
samueljmello: @nogorilla Can you speak to this?

crossroads.net/app/streaming/streamspot.service.ts
samueljmello: Ok. I will take on moving these out to env variables and work with @nogorilla to get the key changed.

crossroads.net/app/downgrades.ts
danrye: Yeah, @silbermm and I talked, we'll leave these in one file so when we are done with Angular 1.5 it'll be isolated.

samueljmello:
@silbermm Unit tests are a separate story (and now branch). @danrye what are your thoughts?

crossroads.net/app/boot.ts
danrye: @samueljmello if that works, lets do it.  I thought we tried that and had issues.

crossroads.net/app/streaming/schedule.component.ts
nogorilla: Correct, Observables are more versatile and reflect the "Angular 2" way and should be prefered going forward. 

In this instance, I initially used Promises; However, after researching Observables vs Promises I have subsequently refactored the StreamspotService to use Observables.

crossroads.net/app/streaming/schedule.component.ts
samueljmello: @dtkocher scratch my comment. I misunderstood what you were asking. Yes, templateUrl was the route we wanted to go. I will change the other references to template to match that. Thanks @danrye

crossroads.net/app/streaming/schedule.component.ts
danrye: That'd be nice if we can bootstrap HTTP_PROVIDERS and then not have to use it all over the place.  I didn't realize you could do that.

crossroads.net/app/streaming/schedule.component.ts
samueljmello: @nogorilla is that in a separate branch?

crossroads.net/app/streaming/streamspot.service.ts
nogorilla: @samueljmello I created a new user story for this. I tried to add the values to `webpack-dev.config.js` and use similar to other environmental variables but that config file doesn't not appear to be compiling for typescript files.

https://rally1.rallydev.com/#/41662702253/detail/userstory/59268553287

samueljmello: I will look into this.

crossroads.net/app/streaming/schedule.component.ts
nogorilla: @samueljmello yes, the refactor is part of US4345 - https://rally1.rallydev.com/#/41662702253/detail/userstory/58514262017

crossroads.net/app/streaming/streamspot.service.ts
danrye: I think this key is a public key... I mean it'll be in the JS file anyway... 

However, what @silbermm said is still true if we are going to have multiple StreamSpot env (test, demo, prod)...

For now I think we can leave it, but need to confirm if it is public or not.

silbermm: @danrye and I just talked... he pointed out to me that no matter what, this key will be in the JS available to the world. I think this still may need to be an env var if it's a different key per environment. No need to change though.

crossroads.net/app/streaming/schedule.component.ts
mjvcallibrity: No objections here.

crossroads.net/app/boot.ts
richardbarker: It works.  Change is in

PR #2375, opened 2016-07-19
DE1523 update process

k8tied1:
pdate process to use participant as trigger field to workaround MP bug


CI/SQL/20160719_133505_DE1523-UpdateJoinGroupNotificationProcessTriggerFieldToParticipentID.sql
```diff
@@ -0,0 +1,7 @@
+use [MinistryPlatform]
+
```
charlie-retzler: You need a GO between the using statement and the update statement

k8tied1: D'oh
added

PR #2380, opened 2016-07-19
US4290: Streamspot Player

tcmacdonald:
This PR delivers [US4290](https://rally1.rallydev.com/#/41662702253d/detail/userstory/58485713334) and [US4387](https://rally1.rallydev.com/#/41662702253d/detail/userstory/58837980731).

Couple notes...

1). The active viewer element at the top of the streaming video page is not saving data currently. That functionality will be addressed via [US4289](https://rally1.rallydev.com/#/41662702253d/detail/userstory/58485711911)

2). The video player is currently using the video.js default stylesheet. We'll address styles for this element via [US4356](https://rally1.rallydev.com/#/41662702253d/detail/userstory/58515804957). 

Let me know if you have any questions. Thanks!


mjvcallibrity:
@tcmacdonald: Everything looks great and there appears to be no TeamCity conflicts. But I ask the dreaded question: `schedule.component` && `streaming.component` && `video.component` && `videojs.component` test coverage status?

Also, I noticed on our html files that we are prefixing component html files with `ng2`. I remember Sam addressing this issue, but I was wondering if this was a convention that needs to be followed in our `component.ts` files as well? My guess is probably not, but I thought I'd ask about this.

samueljmello:
In regards to the ng2 reference in the TS files; I don't think it's necessary. At some point when we fully switch to ng2 we will probably want to remove those references. If we do both the TS and HTML that's twice the amount of files to rename later. So I'm in favor of only doing the HTML templates since that's the only thing that requires it.

tcmacdonald:
Thanks for the feedback. 

I’m still getting my head around testing these components (re: [4345](https://rally1.rallydev.com/#/41662702253d/detail/userstory/58514262017)) but I specifically did not include them here since there’s no real business logic in either of these templates; what little data-binding there is in `video.component.ts` will be replaced with later stories anyway. Short of that, we'd be testing that HTML is getting rendered which I presume is already tested via the framework itself.

Regarding naming– I don't have much of an opinion. I'm inclined to agree w/ Sam but I'd be happy to handle this however if you guys come to a consensus. 

Thoughts?

samueljmello:
I have added unit tests for the new components.

PR #2394, opened 2016-07-20
DE1634: copy corrected

Doug-Shannon:


crossroads.net/app/group_tool/services/createGroup.service.js
```diff
@@ -243,17 +243,17 @@ export default class CreateGroupService {
                     key: 'group.meeting.childcare',
```
JoeKer1: group.kidFriendly

PR #2426, opened 2016-07-22
Feature/us4136 notify of cancelled childcare

andycanterbury:


CI/SQL/20160720_101506_US4136-StoredProcForChildcareCancelledNotifications.sql
```diff
@@ -0,0 +1,40 @@
+SET ANSI_NULLS ON
```
plachmann: Need USE MinistryPlatform

PR #2427, opened 2016-07-24
Defect de1662 edit childcare slow

plachmann:


Gateway/MinistryPlatform.Translation/Repositories/MinistryPlatformRestRepository..cs
```diff
@@ -66,6 +66,32 @@ public List<List<T>> GetFromStoredProc<T>(string procedureName, Dictionary<strin
...
+
+            return (int)response.ResponseStatus;
+        }
+
+        private static string FormatStoredProcBody(Dictionary<string, object> parameters)
+        {
+            string body = "";
+            foreach (var item in parameters)
+            {
+                body += "{\"" + item.Key + "\":\"" + item.Value + "\"},";
```
silbermm: I think this will break when there are multiple parameters... What would happen currently is:

```
{ "key1": "value1"},
{"key2"; "value2}"
```

but what needs to happen is:

```{ "key1": "value1", "key2": "value2"} ```

File: Gateway/MinistryPlatform.Translation/Repositories/MinistryPlatformRestRepository..cs [hunk 2]
```diff
@@ -66,6 +66,32 @@ public List<List<T>> GetFromStoredProc<T>(string procedureName, Dictionary<strin
...
+            }
+           
+            return body;
+            //return "{\"@ChildcareRequestID\":\"184\"}";
```
silbermm: remove old commented code

Gateway/MinistryPlatform.Translation/Repositories/MinistryPlatformRestRepository..cs
```diff
@@ -89,6 +115,11 @@ public List<T> Search<T>(string searchString = null, string selectColumns = null
             return content;
         }
 
+        private string FormatDeleteInts(int[] keys)
```
silbermm: I don't see this method being used anywhere... is it still needed?

PR #2437, opened 2016-07-26
DE1570: Don't init until after state change success so

k8tied1:
Fix bug so that it works the second time


crossroads.net/app/trips/tripgiving/tripgiving.controller.js
```diff
@@ -49,8 +49,8 @@
       GiveTransferService.processing = true;
 
       if (!vm.dto.initialized || toState.name === 'tripgiving') {
-        event.preventDefault();
-        initDefaultState();
+        //event.preventDefault();
+       // initDefaultState();
         return;
```
IngageRitterS: Pleae remove commented code

PR #2438, opened 2016-07-26
US4560: Change country to a drop down list.

JoeKer1:


crossroads.net/app/group_tool/services/createGroup.service.js
```diff
@@ -154,9 +155,12 @@ export default class CreateGroupService {
                     }
                 }, {
                     key: 'profile.address.country',
-                    type: 'input',
+                    type: 'select',
                     templateOptions: {
-                        label: 'Country'
+                        label: 'Country',
+                        valueProp: 'dp_RecordName',
```
sseissiger: Should the valueProp be dp_RecordId instead of dp_RecordName?

PR #2463, opened 2016-07-28
US4559 - Childcare Group Updates

plachmann:


Gateway/Crossroads.ChildcareGroupUpdates/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            TlsHelper.AllowTls12();
+
+            var eventService = container.Resolve<EventService>();
+            var groupService = container.Resolve<GroupService>();
+            var userApiService = container.Resolve<ApiUserRepository>();
+            var mpRestRepository = container.Resolve<MinistryPlatformRestRepository>();
+
+            // use childcare grouptype and eventtype
+            var groupTypeId = 27;
```
silbermm: shouldn't we get these id's from the config?

PR #2481, opened 2016-08-01
US4464: Environment Variables Config

samueljmello:


silbermm:
Looks good. This is something that has been needed for a while. Thanks!

PR #2489, opened 2016-08-02
Defect/DE1688 watch now refresh

nogorilla:


richardbarker:
I noticed that the countdown seems to be somewhat slow in loading.  When I run it I see that the getEvents() call is happening 3 times, and since it pulls all events, each is taking a second or so.  I wonder if we can do something to cache the return value and return it  - or somehow just return the Observable.  

Alternatively, is there any way to just ask streamspot for the next event?

Also, I pulled the development branch into this branch, which seems to have fixed the broken test from yesterday.

nogorilla:
@richardbarker I updated the StreamspotService as a provider for the entire App, that way it is instantiated only once and shared among all other components.

PR #2504, opened 2016-08-03
Fix to show the right button when one processes is happening.

dtkocher:
https://rally1.rallydev.com/#/27593764268d/detail/defect/60295445036


jpc11623:
Reviewed, looks good.

PR #2516, opened 2016-08-04
Feature/us4366 student ministry group email

JoeKer1:


Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -484,5 +484,8 @@
   <!-- Bulk Email Config Settings -->
   <add key="BulkEmailApiUrl" value="https://us12.api.mailchimp.com/3.0/" />
   
+  <!-- Group Tool Config -->
+  <add key="NewStudentMinistryGroupAlertEmailTemplate" value ="280854"/>
```
IngageKrizJ: Should add this to the AsyncJobs MinistryPlatform.config as well.  Might not be needed now, but we try to keep these in sync just in case.

Gateway/crds-angular/Services/GroupService.cs
```diff
@@ -112,6 +114,11 @@ public GroupDTO CreateGroup(GroupDTO group)
                 //save group attributes
                 var configuration = MpObjectAttributeConfigurationFactory.Group();            
                 _objectAttributeService.SaveObjectAttributes(group.GroupId, group.AttributeTypes, group.SingleAttributes, configuration);
+
+                if (group.MinorAgeGroupsAdded)
+                {
+                    _mpGroupService.SendNewStudentMinistryGroupAlertEmail((List<MpGroupParticipant>) mpGroup.Participants);
```
IngageKrizJ: Is this email supposed to go to everyone, or just leaders?

Gateway/crds-angular/MinistryPlatform.config
JoeKer1: Ahh, I forgot to add them! I'll commit the new values and double check any other ones we've added recently..

Gateway/crds-angular/Services/GroupService.cs
JoeKer1: It is actually an email that goes to student ministry with the names and email address of the leaders.

IngageKrizJ: And when creating a group, there is only one participant (the leader), so this is ok.  Thanks for clarifying!

PR #2517, opened 2016-08-04
US4536 Add API/support for scheduling email to be sent at specific datetime

richardbarker:


charlie-retzler:
PR needs Labels and Milestones configured

Gateway/crds-angular/Controllers/API/EmailCommunicationController.cs
```diff
@@ -58,5 +58,22 @@ public IHttpActionResult Post([FromBody] CommunicationDTO communication)
                 }
             });
         }
+
+        /// <summary>
+        /// Schedule an email to a specific contactId/emailAddress at a specific time
+        /// </summary>
+        [Route("api/scheduleemail")]
+        public IHttpActionResult PostReminder(EmailCommunicationDTO email)
```
charlie-retzler: This method seems like a duplicate of the method below with the exception of Authentication. Also, how is this protected from misuse/abuse?
 
[Route("api/sendemail")] public IHttpActionResult Post(EmailCommunicationDTO email)

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
...
+            if (token == null && email.FromUserId == null && template.FromContactId == 0)
             {
-                throw (new InvalidOperationException("Must provide either email.FromUserId or an authentication token."));
+                throw (new InvalidOperationException("Must provide either email.FromUserId from  or an authentication token."));
             }
 
-            var replyToContactId = email.ReplyToContactId ?? DefaultContactEmailId;
-            var from = new MpContact { ContactId = DefaultContactEmailId, EmailAddress = _communicationService.GetEmailFromContactId(DefaultContactEmailId) };
+            //ReplyToContact - If passed in, use it.  Otherwise, get from template.
+            var replyToContactId = email.ReplyToContactId ?? template.ReplyToContactId;
```
charlie-retzler: Is this change in functionality where DefaultContactEmailId is no longer used? Was any code relying on this being DefaultContactEmailId?

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
...
 
-            var template = _communicationService.GetTemplate(email.TemplateId);
-            communication.TemplateId = email.TemplateId;
-            communication.EmailBody = template.Body;
-            communication.EmailSubject = template.Subject;
+            //ToContacts - If passed in, use it.  Otherwise, get from emailAddress.  If no contact for the emailAddress, use the default.
+            MpContact to;
+            if (email.ToContactId != null)
+            {
+                to = new MpContact { ContactId = (int)email.ToContactId, EmailAddress = _communicationService.GetEmailFromContactId((int)email.ToContactId) };
```
charlie-retzler: When working with nullable types don't cast, to int use email.ToContactId.Value.

Also may consider switching check above to email.ToContactId.HasValue() for nullable type

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
...
+            //FromContact - If passed in, use it.  Otherwise, get from template.
+            var fromContactId = email.FromContactId ?? template.FromContactId;
+            var from = new MpContact { ContactId = DefaultContactEmailId, EmailAddress = _communicationService.GetEmailFromContactId(fromContactId) };
 
-            var template = _communicationService.GetTemplate(email.TemplateId);
-            communication.TemplateId = email.TemplateId;
-            communication.EmailBody = template.Body;
-            communication.EmailSubject = template.Subject;
+            //ToContacts - If passed in, use it.  Otherwise, get from emailAddress.  If no contact for the emailAddress, use the default.
+            MpContact to;
```
charlie-retzler: I would suggest extracting the logic for getting the MPContact To field to a seperate function to make code more readable and more maintainable.

Gateway/crds-angular/Controllers/API/EmailCommunicationController.cs
```diff
@@ -58,5 +58,22 @@ public IHttpActionResult Post([FromBody] CommunicationDTO communication)
...
+        {
+            try
+            {
+                _emailCommunication.SendEmail(email);
+                return this.Ok();
+            }
+            catch (System.Exception ex)
+            {
+                return this.InternalServerError();
```
charlie-retzler: We should probably log the error here, before we loose the stack and specific error message.

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
...
+                catch (Exception)
+                {
+                    //work around incorrectly handled case where multiple contacts exists for a single contact
+                    contactId = 0;
+                }
 
-            communication.MergeData = email.MergeData;
+                if (contactId == 0)
+                {
+                    contactId = DefaultContactEmailId;
```
charlie-retzler: Can this logic be simplified by moving DefaultContactEmailId to the "var contactId = 0;"?

Gateway/crds-angular/Controllers/API/EmailCommunicationController.cs
danrye: On the misuse/abuse we are open to other ideas, BST has another story on their backlog for securing unauthed endpoints... my only idea at the moment is client keys tied to domains, or using some sort of HMAC.  That is why I started looking into JWT.  Nothing is fullproof with client keys it seems.

The quick fix I've suggested is to limit this endpoint to only that one template.  At the moment the SendEmail method doesn't support custom bodies, it has to use a template for the content it seems, so the misuse is somewhat limited... And as for DDS type things, that's where the WAF should come into play.  Anyway... open to ideas here.

danrye: @charlie-retzler ^

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
...
+            //ReplyToContact - If passed in, use it.  Otherwise, get from template.
+            var replyToContactId = email.ReplyToContactId ?? template.ReplyToContactId;
             var replyTo = new MpContact { ContactId = replyToContactId, EmailAddress = _communicationService.GetEmailFromContactId(replyToContactId) };
-            var recipient = new MpContact {ContactId = email.ToContactId, EmailAddress = _communicationService.GetEmailFromContactId(email.ToContactId) };
 
-            communication.FromContact = from;
-            communication.ReplyToContact = replyTo;
-            communication.ToContacts.Add(recipient);
+            //FromContact - If passed in, use it.  Otherwise, get from template.
+            var fromContactId = email.FromContactId ?? template.FromContactId;
```
danrye: Hmm, I think we run into some trouble here... if FromContactId isn't supplied, you want to use the templates FromContactId... which is the right thing to do... but it used to use DefaultContactEmailId.  I'm not sure what to do here, your doing the right thing, but I'm sure we've depended on the wrong thing here for awhile...

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,65 @@ public class EmailCommunication : IEmailCommunication
 
         public void SendEmail(EmailCommunicationDTO email, string token)
```
charlie-retzler: Are Dev/QA aware of what functionality is impacted by this change? Do they have capacity to do regression test those parts of the application?

danrye: It isn't going to make this sprint, so next sprint the QA effort will need to be factored in.

Gateway/crds-angular/Services/EmailCommunication.cs
richardbarker: I did change this from DefaultContactEmailId to the template contact.  Since template is required, it seemed like the correct thing to do.  Please let me know if that is incorrect and I'll set it accordingly.

Gateway/crds-angular/Services/EmailCommunication.cs
richardbarker: I'm trying to correctly handle the case where GetContactIdByEmail fails.  Since GetContactIdByEmail will return 0 if no contacts are found, or throw an exception if more than one are found, there are 2 cases to handle.  I have to check if the return contactId == 0, even if I initialize it to DefaultContactEmailId.  Am I missing something?

Gateway/crds-angular/Services/EmailCommunication.cs
danrye: Yep, this is the right thing to do.  It is essentially fixing a bug.

danrye:
Ready to Merge, but don't want to merge until after the demo deploy

Gateway/crds-angular.test/Services/EmailCommunicationTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            _communicationService.Setup(mocked => mocked.GetEmailFromContactId(template.ReplyToContactId)).Returns(template.ReplyToEmailAddress);
+            _communicationService.Setup(mocked => mocked.GetEmailFromContactId(template.FromContactId)).Returns(template.FromEmailAddress);
+            _communicationService.Setup(mocked => mocked.GetTemplate(emailData.TemplateId)).Returns(template);
+            _contactService.Setup(mocked => mocked.GetContactIdByEmail(emailData.emailAddress)).Returns(10);
+
+            _communicationService.Setup(m => m.SendMessage(It.IsAny<MpCommunication>(), false)).Verifiable();
+            try
+            {
+                fixture.SendEmail(emailData, null);
+                _communicationService.Verify(m => m.SendMessage(It.IsAny<MpCommunication>(), false), Times.Once);
```
charlie-retzler: Can this verify that it is being sent to the correct contact?

Gateway/crds-angular.test/Services/EmailCommunicationTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            _communicationService.Setup(mocked => mocked.GetEmailFromContactId(emailData.ToContactId.Value)).Returns("user@test.com");
+            _communicationService.Setup(mocked => mocked.GetEmailFromContactId(template.FromContactId)).Returns(template.FromEmailAddress);
+            _communicationService.Setup(mocked => mocked.GetTemplate(emailData.TemplateId)).Returns(template);
+            _contactService.Setup(mocked => mocked.GetContactIdByEmail(emailData.emailAddress)).Returns(0);
+
+            _communicationService.Setup(m => m.SendMessage(It.IsAny<MpCommunication>(), false)).Verifiable();
+            try
+            {
+                fixture.SendEmail(emailData, null);
+                _communicationService.Verify(m => m.SendMessage(It.IsAny<MpCommunication>(), false), Times.Once);
```
charlie-retzler: Can this verify that it is being sent to the correct contact?

Gateway/crds-angular.test/Services/EmailCommunicationTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            }
+            catch (Exception)
+            {
+                Assert.Fail("Unexpected exception was thrown");
+            }
```
charlie-retzler: Do we need these try / catches here? A test that throws an exception without an ExpectedException should already fail and give you stack trace information.

Gateway/crds-angular/Controllers/API/EmailCommunicationController.cs
charlie-retzler: Are we putting anything in place to limit template like Dan referenced above? Also is the naming of this route/method correct since it can be used for scheduled emails or really any unauthenticted unscheduled emails?

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,38 @@ public class EmailCommunication : IEmailCommunication
...
-            var communication = new MpCommunication();
-            communication.DomainId = DomainID;
-            communication.AuthorUserId = email.FromUserId ?? DefaultAuthorUserId;
+            var template = _communicationService.GetTemplate(email.TemplateId);
 
-            if (token == null && email.FromUserId == null)
+            if (token == null && email.FromUserId == null && template.FromContactId == 0)
             {
-                throw (new InvalidOperationException("Must provide either email.FromUserId or an authentication token."));
+                throw (new InvalidOperationException("Must provide either email.FromUserId from  or an authentication token."));
```
charlie-retzler: It looks like the word from was added to error message, but doesn't make a ton of sense

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,38 @@ public class EmailCommunication : IEmailCommunication
...
+            var replyToContactId = email.ReplyToContactId ?? template.ReplyToContactId;
             var replyTo = new MpContact { ContactId = replyToContactId, EmailAddress = _communicationService.GetEmailFromContactId(replyToContactId) };
-            var recipient = new MpContact {ContactId = email.ToContactId, EmailAddress = _communicationService.GetEmailFromContactId(email.ToContactId) };
 
-            communication.FromContact = from;
-            communication.ReplyToContact = replyTo;
-            communication.ToContacts.Add(recipient);
+            //FromContact - If passed in, use it.  Otherwise, get from template.
+            var fromContactId = email.FromContactId ?? template.FromContactId;
+            var from = new MpContact { ContactId = DefaultContactEmailId, EmailAddress = _communicationService.GetEmailFromContactId(fromContactId) };
```
charlie-retzler: Should the ContactId = DefaultContactEmailId be ContactId = fromContactId here?

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -35,30 +35,38 @@ public class EmailCommunication : IEmailCommunication
...
 
-            var template = _communicationService.GetTemplate(email.TemplateId);
-            communication.TemplateId = email.TemplateId;
-            communication.EmailBody = template.Body;
-            communication.EmailSubject = template.Subject;
-
-            communication.MergeData = email.MergeData;
+            //ToContacts - If passed in, use it.  Otherwise, get from emailAddress.  If no contact for the emailAddress, use the default.
+            MpContact to = GetMpContactFromEmailCommunicationDto(email);
```
charlie-retzler: Naming of method here is confusing. The method is names like it is getting the FromEmail, but is being assigned to the ToEmail.

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -68,6 +76,38 @@ public void SendEmail(EmailCommunicationDTO email, string token)
...
+                }
+                if (contactId == 0)
+                {
+                    contactId = DefaultContactEmailId;
+                }
+                to = new MpContact { ContactId = contactId, EmailAddress = email.emailAddress };
+            }
+            else
+            {
+                throw (new InvalidOperationException("Must provide either ToContactId or emailAddress."));
```
charlie-retzler: I would prefer this to be the first thing in the method rather than if / then / else nesting.

Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -68,6 +76,38 @@ public void SendEmail(EmailCommunicationDTO email, string token)
...
+        private MpContact GetMpContactFromEmailCommunicationDto(EmailCommunicationDTO email)
+        {
+            MpContact to;
+            if (email.ToContactId.HasValue)
+            {
+                to = new MpContact { ContactId = email.ToContactId.Value, EmailAddress = _communicationService.GetEmailFromContactId(email.ToContactId.Value) };
+            }
+            else if (email.emailAddress != null)
+            {
+                var contactId = 0;
```
charlie-retzler: May be able to clean this up a little bit, but would be easier to talk through and decide if we want to.

Gateway/crds-angular.test/Services/EmailCommunicationTest.cs
```diff
@@ -0,0 +1,170 @@
...
+            _contactService.Setup(mocked => mocked.GetContactIdByEmail(emailData.emailAddress)).Returns(0);
+
+            var spiedComm = new MpCommunication();
+            _communicationService.Setup(m => m.SendMessage(It.IsAny<MpCommunication>(), false))
+                .Callback<MpCommunication, bool>((comm, token) => spiedComm = comm);
+
+            
+            fixture.SendEmail(emailData, null);
+            _communicationService.Verify(m => m.SendMessage(It.IsAny<MpCommunication>(), false), Times.Once);
+            Assert.AreEqual(JsonConvert.SerializeObject(expectedContact),JsonConvert.SerializeObject(spiedComm.ToContacts[0]));
```
charlie-retzler: This is a great solution!

PR #2520, opened 2016-08-05
US4644 - Participant and Requests tab help content

tallented:
* Always show help content for participant
* Show Request help content when there are no inquiries or invites
* Hide participant Edit and Email Group buttons when leader is the only person in group (count == 1)

image

image


dtkocher:
👍 Looks good and will merge when build passes.

PR #2556, opened 2016-08-10
Feature/us4589 past weekends cms

nogorilla:


crossroads.net/app/streaming/streaming.component.ts
```diff
@@ -34,6 +53,31 @@ export class StreamingComponent {
...
+
+              event.image = 'https://crds-cms-uploads.imgix.net/content/images/register-bg.jpg'
+              if (typeof event.messageVideo.still !== 'undefined') {
+                event.image = event.messageVideo.still.filename
+              } 
+              event.imageSrc = event.image.replace(/https*:/, '')
+
+              this.cmsDataService.getSeries(`id=${event.series}`)
+                .subscribe((series) => {
+                  event.seriesTitle = series.length > 0 ? _.first(series).title : 'Message';
```
mjvcallibrity: Is there another way to accomplish this functionality? Seeing the nested `.subscribe()` is throwing a red flag, not because I actually know better, but I would imagine there is a better pattern out there. [Perhaps something like this?](https://www.reddit.com/r/Angular2/comments/4cgp1d/handle_nested_observable_subscribes/)

crossroads.net/app/streaming/streaming.component.ts
```diff
@@ -34,6 +53,31 @@ export class StreamingComponent {
...
+        // only 4 are displayed, however, there is an invalid message on contentint that no one can seem to find,
+        //   so grabing an extra message and then filtering out the bad message in the view
+        .getXMostRecentMessages(5)
+        .subscribe((mostRecent) => {
+          this.mostRecent = mostRecent;
+
+          _.each(mostRecent, (event, key) => {
+            if (typeof event.series !== "undefined") {
+              event.delay = key * 100;
+
```
mjvcallibrity: It isn't immediately obvious to me what `event.delay` is and what is function is. Would it be possible to get an explanation?

nogorilla: delay is a value used by WOW.js to determine how long to wait before showing the content. The old layout shows that in effect: https://github.com/crdschurch/crds-angular/pull/2556/files/737c32327c58d5d6e7fa16baeff3c624591e3237#diff-0f677f720b17fc8167b2b7b744fd590aL49

PR #2586, opened 2016-08-15
US4696 - Request to Join - Confirm Request Modal

tallented:
Note:   UI layout only, Submitting request functionality is mocked
Clicking "Back to Search Results" dismisses the modal
Clicking on the darkened window overlay dismisses the modal
Clicking the "Send Request" button mocks the API call with 2 sec timeout, then dismisses modal

### Displays as large modal on Desktop
image

### Displays as small modal on smaller screens.  
"Description" label stacks with description text to avoid overlaying
image

### Displays as scrollable modal on mobile screens
image

### Displays correctly on older iPhones
image

### Clicking "Send Request" displays button loading state "Sending..."
image

### DIFF
--- a/crossroads.net/app/group_tool/group_search_results/groupSearchResults.controller.js
+++ b/crossroads.net/app/group_tool/group_search_results/groupSearchResults.controller.js
@@ -19,10 +19,10 @@ export default class GroupSearchResultsController {
 
   $onInit() {
     this.search = {
-      query: this.state.params.query || 'tallent',
+      query: this.state.params.query,
       location: this.state.params.location
     };
-    this.doSearch(this.state.params.query || 'tallent', this.state.params.location);
+    this.doSearch(this.state.params.query, this.state.params.location);
   }
 
   doSearch(query, location) {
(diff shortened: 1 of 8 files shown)


crossroads.net/app/group_tool/group_search_results/groupSearchResults.controller.js
```diff
@@ -60,4 +61,22 @@ export default class GroupSearchResultsController {
   submit() {
     this.doSearch(this.search.query, this.search.location);
   }
+
+  requestToJoin(group) {
+    console.debug("Request to Join", group);
+    var modalInstance = this.$modal.open({
+      template: '<confirm-request group="confirmRequestModal.group" modal-instance="confirmRequestModal.modalInstance"></confirm-request>',
+      controller: function(group, $modalInstance) {
```
tallented: I'm not sure here if having an inline controller is the correct way to handle this.  I did this in order to pass the group and modal instance to the actual Confirm Request component.   The controller and template here could be externalized to their own files?

PR #2587, opened 2016-08-15
Feature/us4648 refactor unit test group tool

ilpadre:


crossroads.net/app/group_tool/services/createGroup.service.js
```diff
@@ -768,25 +784,47 @@ export default class CreateGroupService {
...
+    mapToSmallGroup() {
+        let smallGroup = new SmallGroup();
+        this.mapToSmallGroupAbout(smallGroup);
+        this.mapToSmallGroupType(smallGroup);
+        this.mapToSmallGroupSingleAttributes(smallGroup);
+        this.mapToSmallGroupMultipleAttributes(smallGroup);
+        this.mapToSmallGroupMeetingDay(smallGroup);
+        this.mapToSmallGroupMeetingTime(smallGroup);
+        this.mapToSmallGroupMeetingPlace(smallGroup);
+        this.mapToSmallGroupCategory(smallGroup);
```
Doug-Shannon: I'm not in love with passing in the small group object and setting data on it without returning anything.  Not that it is wrong, it can just be hard to follow.  I think it would be a little more descriptive to pass small group into the function and use it to return some piece of data that is set to small group on the calling side.  It may just be a matter of preference.

PR #2589, opened 2016-08-15
Fix script to handle weird characters.

dtkocher:


jpc11623:
Reviewed, looks good.

PR #2617, opened 2016-08-17
US4568: Add required message functionality to message component

IngageKrizJ:
* [US4568 - Require message when denying request or removing participant](https://rally1.rallydev.com/#/27593764268d/detail/userstory/59861948287)
* Note that this story is only to refactor the message component to allow users to specify whether the message is required or not, and prevent sending an email without a message if required.
* There is an existing defect, [DE1775](https://rally1.rallydev.com/#/27593764268d/detail/defect/60974721637), that will address issues with the actual email.

### DIFF
--- a/crossroads.net/app/group_tool/group_message/groupMessage.controller.js
+++ b/crossroads.net/app/group_tool/group_message/groupMessage.controller.js
@@ -1,9 +1,26 @@
 
 export default class GroupMessageController {
   /*@ngInject*/
-  constructor() { }
+  constructor($rootScope) {
+    this.rootScope = $rootScope;
+  }
 
-  submit() {
+  $onInit() {
+    this.emailMessageRequired = Boolean(this.emailMessageRequired);  
+    if(!this.emailMessageLabel) {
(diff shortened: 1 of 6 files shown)


crossroads.net/app/group_tool/group_message/groupMessage.controller.js
```diff
@@ -1,9 +1,26 @@
 
 export default class GroupMessageController {
   /*@ngInject*/
-  constructor() { }
+  constructor($rootScope) {
+    this.rootScope = $rootScope;
+  }
+
+  $onInit() {
+    this.emailMessageRequired = Boolean(this.emailMessageRequired);
```
dtkocher: @IngageKrizJ Does this make undefined into false?

IngageKrizJ: @dtkocher - Yep.  I originally wrote it with the whole undefined/null check, but thought there might be a better way.  Sure enough, there was, and I put tests in place for it as well.

PR #2631, opened 2016-08-17
Defect/de1775 emails not sending on approve and deny

IngageKrizJ:
* [DE1775 - Approve/Deny/Remove Emails not working](https://rally1.rallydev.com/#/27593764268d/detail/defect/60974721637)
* Two issues with this defect:

1. Email template from CMS was not being parsed
 * Since we pull from CMS, we have to replace tokens in CMS content block BEFORE calling communication service to send email
2. Sending approve/deny for inquiry was not working 
 * Since an inquiry is not a group participant yet, logic to get participant from group was not working - needed to lookup participant from MP instead


jpc11623:
Reviewed, everything is awesome.

PR #2634, opened 2016-08-18
Defect/de1802 filters not resetting

IngageKrizJ:
* [DE1802 - Handle reset of filters when clicking cancel](https://rally1.rallydev.com/#/27593764268d/detail/defect/61127475044)
* Also do some minor code cleanup to remove console.log() calls, and some window.setTimeout() calls that are no longer needed


jpc11623:
Reviewed, this also looks awesome.

PR #2669, opened 2016-08-19
Feature/us4583 save stream reminder text

richardbarker:
Please consult with me before merging this.  We need to determine if we have enough time to adequately test this before the deadline.


Gateway/Crossroads.AsyncJobs.Test/Crossroads.AsyncJobs.Test.csproj
```diff
@@ -65,13 +73,45 @@
...
     <Reference Include="Moq">
       <HintPath>..\packages\Moq.4.2.1502.0911\lib\net40\Moq.dll</HintPath>
     </Reference>
     <Reference Include="nunit.framework">
       <HintPath>..\packages\NUnit.2.6.4\lib\nunit.framework.dll</HintPath>
     </Reference>
+    <Reference Include="Quartz, Version=2.3.3.0, Culture=neutral, PublicKeyToken=f6b8c98a402cc8a4, processorArchitecture=MSIL">
+      <HintPath>..\packages\Quartz.2.3.3\lib\net40\Quartz.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
```
IngageKrizJ: Assume this is needed by quartz? We already have log4net, so should use that in our own code.

Gateway/Crossroads.AsyncJobs/Processors/ScheduledJobProcessor.cs
```diff
@@ -0,0 +1,34 @@
+﻿using crds_angular.Models.Crossroads;
+using Crossroads.AsyncJobs.Interfaces;
+using log4net;
+using Quartz;
+using Quartz.Impl;
+
+namespace Crossroads.AsyncJobs.Processors
+{
+    public class ScheduledJobProcessor : IJobExecutor<ScheduledJob>
```
IngageKrizJ: Does this processor pick up messages to schedule jobs in quartz? I thought we talked that we were not going to do that, unless there was a real need to have the scheduling happen asynchronously (which there may be...)

Gateway/Crossroads.AsyncJobs.Test/Crossroads.AsyncJobs.Test.csproj
richardbarker: Yep, dependency of Quartz.

Gateway/Crossroads.AsyncJobs/Processors/ScheduledJobProcessor.cs
richardbarker: The processor does pick up messages to schedule jobs in Quartz.  I thought we said having it in AsyncJobs can be beneficial because it is (supposed to be) up consistently.  I left it there based on that, but you're right there is no explicit need for it to be there.

richardbarker:
Need to move Twilio account info into environment variables.

Gateway/Crossroads.AsyncJobs/MinistryPlatform.Demo.config
```diff
@@ -5,4 +5,12 @@
   <!-- Configuration values for Stripe Event Handler (webhooks) -->
   <add key="StripeWebhookAsynchronousProcessingMode" value="true" xdt:Transform="Replace" xdt:Locator="Match(key)"/>
   <add key="CheckScannerDonationsAsynchronousProcessingMode" value="true" xdt:Transform="Replace" xdt:Locator="Match(key)"/>
+
+  <!-- Configuration Values for Texts -->
+  <add key="ScheduledJobsQueueName" value=".\Private$\ScheduledJobs"/>
+  <add key="TwilioFromPhoneNumber" value="+15005550006"/>
+  <add key="TwilioAccountSid" value="AC051651a7abfd7ec5209ad22273a24390"/>
+  <add key="TwilioAuthToken" value="4995535b191fe6e7dd3578c10e4c7976"/>
+  <add key="StreamReminderTemplate" value="264567"/>
```
Doug-Shannon: Is there a better place for these values?

Gateway/crds-angular/job_scheduling_data_2_0.xsd
```diff
@@ -0,0 +1,361 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
```
Doug-Shannon: This file appears 4 times throughout the PR in different locations.  Is it completely duplicated or subtly different?  If it is the same, is there a way to reuse it?

Gateway/Crossroads.AsyncJobs/MinistryPlatform.Demo.config
IngageKrizJ: They should be environment variables, assuming they are private/secret keys for Twilio.  (Also would need to be changed now that they are out in public github repo)

PR #2675, opened 2016-08-22
Feature/us4583 save stream reminder text

richardbarker:


Gateway/Crossroads.AsyncJobs/Processors/ScheduledJobProcessor.cs
```diff
@@ -0,0 +1,34 @@
+﻿using crds_angular.Models.Crossroads;
+using Crossroads.AsyncJobs.Interfaces;
+using log4net;
+using Quartz;
+using Quartz.Impl;
+
+namespace Crossroads.AsyncJobs.Processors
+{
+    public class ScheduledJobProcessor : IJobExecutor<ScheduledJob>
```
IngageKrizJ: Rick, let's talk tomorrow, I'd still like to understand if we can get rid of this and just make calls remotely to schedule jobs.  I honestly don't know what options are available (if any), so if there isn't some easily exposed REST (or other) API for Quartz, then this may be the way to go.

Gateway/crds-angular/Controllers/API/TextCommunicationController.cs
```diff
@@ -0,0 +1,89 @@
...
+            _messageQueue = messageQueue;
+            _messageQueue.CreateQueue(_eventQueueName, QueueAccessMode.Send);
+            _messageFactory = messageFactory;
+        }
+
+        /// <summary>
+        /// Schedule a text message to a specific contactId/number at a specific time
+        /// </summary>
+        [Route("api/sendTextMessageReminder")]
+        public IHttpActionResult PostReminder(TextCommunicationDto textCommunicationData)
```
IngageKrizJ: Should this require authentication?  I vaguely remember this being talked about last sprint, but can't remember the final verdict.

Gateway/crds-angular/Models/Crossroads/ScheduledJob.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using Quartz;
+
+namespace crds_angular.Models.Crossroads
+{
+    public class ScheduledJob
+    {
+        public DateTime StartDateTime { get; set; }
+        public Type JobType { get; set; }
+        public Object Dto { get; set; }
```
IngageKrizJ: Kind of nitpicking here, but....  Could the JobType and Dto objects be made generic?  So the ScheduledJob class would actually be something like below.  May not work, since it is being serialized and deserialized from JSON on the queue, but I wanted to at least ask.  Anytime I see a Type and an Object being used like this, I look for ways to use generics in place of that.
```c#
public class ScheduledJob<T> {
    public DateTime StartDateTime { get; set; }
    public T Dto { get; set; }
}
```

Gateway/crds-angular/Services/TwilioService.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public ILog _logger = LogManager.GetLogger(typeof(TwilioService));
+
+        private readonly string _fromPhoneNumber;
+        private readonly TwilioRestClient _twilio;
+
+        public TwilioService(IConfigurationWrapper configurationWrapper)
+        {
+            var accountSid = configurationWrapper.GetConfigValue("TwilioAccountSid");
+            var authToken = Environment.GetEnvironmentVariable("TWILIO_AUTH_TOKEN");
```
IngageKrizJ: The IConfigurationWrapper actually has a `GetEnvironmentVarAsString(string variable)` method you can use here.

Gateway/crds-angular/Services/TwilioService.cs
```diff
@@ -0,0 +1,40 @@
...
+            _fromPhoneNumber = configurationWrapper.GetConfigValue("TwilioFromPhoneNumber");
+            _twilio = new TwilioRestClient(accountSid, authToken);
```
IngageKrizJ: This will make it difficult to unit test.  Is it possible to inject TwilioRestClient from Unity, instead of newing it here?

Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -489,12 +489,16 @@
...
   <!-- Group Tool Config -->
   <add key="NewStudentMinistryGroupAlertEmailTemplate" value ="280854"/>
   <add key="GroupEndedParticipantEmailTemplate" value ="290789"/>
   <add key="AddressMatrixSearchDepth" value="5"/>
 
+  <!-- Configuration Values for Texts -->
+  <add key="ScheduledJobsQueueName" value=".\Private$\ScheduledJobs"/>
+  <add key="StreamReminderTemplate" value="264567"/>
+  <add key="TwilioFromPhoneNumber" value="+15134020289"/>
+  <add key="TwilioAccountSid" value="AC4d2b824e1b80597265566e98f903ac17"/>
```
IngageKrizJ: Is this value allowed to be seen by anyone outside Crossroads?  If not, we should make this an environment variable as well.

Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -489,12 +489,16 @@
...
-  <add key="StreamReminderTemplate" value="264567"/>
   
   <!-- Group Tool Config -->
   <add key="NewStudentMinistryGroupAlertEmailTemplate" value ="280854"/>
   <add key="GroupEndedParticipantEmailTemplate" value ="290789"/>
   <add key="AddressMatrixSearchDepth" value="5"/>
 
+  <!-- Configuration Values for Texts -->
+  <add key="ScheduledJobsQueueName" value=".\Private$\ScheduledJobs"/>
+  <add key="StreamReminderTemplate" value="264567"/>
```
IngageKrizJ: Does this template exist in all environments already?  If not, it should be scripted (I can point you to example scripts, we have kind of a boilerplate script for these now).

Gateway/crds-angular/Models/Crossroads/ScheduledJob.cs
richardbarker: I tried this but it is giving problems where it is trying to deserialize it.  OK if I just leave it as is?

PR #2684, opened 2016-08-25
Feature/us4787 search filter online or in person

IngageKrizJ:
* [US4787 - Group Location Search Filter](https://rally1.rallydev.com/#/27593764268d/detail/userstory/61460348432)
* [US4774 - Kids Welcome Search Filter](https://rally1.rallydev.com/#/27593764268d/detail/userstory/61284109629)
* A UI story ([US4800](https://rally1.rallydev.com/#/27593764268d/detail/userstory/61465760426)) will fix the display of the filters (they are currently horizontal in rows, but will be changed to vertical columns for each filter)


jpc11623:
It looks good, but my one question is how are we making sure the same result doesn't show up in the results grid twice - I'm sure it's in there, but probably just missed it.

PR #2687, opened 2016-08-25
US4802: Allow for current series to fall back

mjvcallibrity:
Allow for current series to fall back to last series if no series is scheduled into the future.


crossroads.net/app/streaming/current-series.component.ts
```diff
@@ -33,22 +33,30 @@ export class CurrentSeriesComponent {
...
+  response: any;
   
   constructor(private cmsDataService: CMSDataService) {
   }
   
   ngOnInit() {
     this.cmsDataService.getCurrentSeries().subscribe((response) => {
       this.parseData(response);
+      if (this.response == undefined) {
+        this.cmsDataService.getLastSeries().subscribe((response) => {
```
richardbarker: This should probably be:
if (this.response === undefined) {
nitpicky

PR #2707, opened 2016-08-26
Feature/us4589 past weekends cms

nogorilla:


crossroads.net/app/streaming/streaming.component.ts
```diff
@@ -84,7 +84,15 @@ export class StreamingComponent {
 
               event.delay = i * 100;
               event.subtitle = event.title
-              event.title = '';
+              event.title = event.series.title;
+
+              if (event.sequence.length > 0) {
+                let index = _.findIndex(event.sequence, {id: event.id});
```
samueljmello: You don't have to do this. I included the "number" property so you wouldn't have to figure this out.

PR #2731, opened 2016-08-30
Defect/de1850 missing frequency private invite

ilpadre:


Gateway/crds-angular.test/Services/GroupServiceTest_REMOTE_16368.cs
```diff
@@ -0,0 +1,996 @@
+﻿using System;
```
JoeKer1: I think this got added by accident

PR #2734, opened 2016-08-30
US3349 - Trip App Confirmation Email

plachmann:


Gateway/crds-angular/Controllers/API/DonationController.cs
```diff
@@ -218,7 +218,7 @@ private IHttpActionResult CreateDonationAndDistributionAuthenticated(String toke
                     Anonymous = dto.Anonymous
                 };
 
-                var donationId = _mpDonorService.CreateDonationAndDistributionRecord(donationAndDistribution);
+                var donationId = _mpDonorService.CreateDonationAndDistributionRecord(donationAndDistribution, !dto.TripDeposit);
```
andycanterbury: I'm confused by the !dto.TripDeposit. If its not a deposit, so scholarship, then we do send an email about the donation? There shouldn't be a donation if its scholarshipped?

PR #2759, opened 2016-08-31
Feature/us3276 add pledge donor to email

andycanterbury:


Gateway/MinistryPlatform.Translation/Repositories/MinistryPlatformRestRepository..cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Policy;
```
silbermm: is this really necessary?

PR #2768, opened 2016-09-01
US2413 - Guest Giver provides first and last name

plachmann:


Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -40,14 +40,16 @@ public ContactRepository(IMinistryPlatformService ministryPlatformService, IAuth
             _participantsPageId = configuration.GetConfigIntValue("Participants");
         }
 
-        public int CreateContactForGuestGiver(string emailAddress, string displayName)
+        public int CreateContactForGuestGiver(string emailAddress, string displayName, string firstName, string lastName)
```
silbermm: you could default these params to empty strings. This way we don't have to pass string.empty or "" to the method when we don't care about the name.

Gateway/crds-angular/Services/DonorService.cs
```diff
@@ -134,7 +138,12 @@ public MpContactDonor CreateOrUpdateContactDonor(MpContactDonor mpContactDonor,
                 }
                 else
                 {
-                    contactDonorResponse.ContactId = _mpContactService.CreateContactForGuestGiver(emailAddress, _guestGiverDisplayName);
+                    var displayName = _guestGiverDisplayName;
+                    if (!string.IsNullOrEmpty(firstName) && !string.IsNullOrEmpty(lastName))
+                    {
+                        displayName = lastName + "," + firstName;
```
silbermm: better to use string interpolation here? `displayName = $"{lastName}, {firstName}`

crossroads.net/app/trips/tripgiving/tripgivingTemplates/account.html
```diff
@@ -33,6 +33,19 @@
   <hr>
 
     <ng-form name="accountForm">
+
+      <div class='row' ng-show='tripGiving.showNameInput()'>
+        <div class="form-group col-sm-6" ng-class="{'has-error': accountForm.input.donorFirstName.$invalid }">
+          <label for='firstName' class='control-label'> First Name </label>
+          <input ng-model="tripGiving.dto.donorFirstName" type='text' class='form-control' name="donorFirstName" id='donor-first-name' placeholder='First Name' >
+        </div>
+
```
silbermm: fields are not required, and should be

PR #2774, opened 2016-09-01
Feature/us4684 submit event reminder data

nogorilla:
This PR also includes fixes for DE1892 and DE1891


crossroads.net/app/streaming/reminder.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(private http: Http) { }
+
+  sendTextReminder(reminder): Promise<any> {
+    let body = JSON.stringify({ 
+      "templateId": 0,
+      "mergeData": {
+        "Event_Date": reminder.day,
+        "Event_Start_Time": reminder.time
+      },
+      "toPhoneNumber": reminder.phone,
```
mjvcallibrity: Twilio requires the phone number to be formatted in the "+12345678912" format. Is there a way that we can transform the input from the form to that format before we make a POST request?

PR #2790, opened 2016-09-07
DE1905 - Set startDate to selected date/time

nogorilla:


crossroads.net/app/streaming/reminder.ts
```diff
@@ -22,6 +23,13 @@ export class Reminder {
 
   public send(): Promise<any> {
     let result = null;
+    let offset = '-0500'; // default to EST offset
+    if (this.time.endsWith('EDT')) {
+      offset = '-0400';
+    }
+    let time = this.time.slice(0, this.time.length - 4); // strip of tz code and leading space
+
+    this.startDate = moment(`${this.day} ${time}${offset}`, 'MM/DD/YYYY h:mma').format();
```
richardbarker: Is this going to be a wrong if you aren't in eastern time zone?  The date options come from the streamspot api, right?  Those should contain timezone info.  Can we somehow preserve that and so we don't have to convert it?

richardbarker:
Looks like StreamSpot does not send any timezone information in the event dates.  We have to hardcode EST for now - to be revisited if/when StreamSpot sends dates like a grown up API.

PR #2822, opened 2016-09-12
US4911 improvements to email handling on profile

andycanterbury:


crossroads.net/app/common/profile/personal/emailChange.service.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  reset() {
+      this.originalEmail = '';
+      this.isSet = false;
+  }
+
+}
+
+export default EmailChangeService;
```
samueljmello: Did you not want to define the properties originalEmail and isSet on the class? I see that the constructor is basically creating them, but you may want to do that on the class itself and set them as private that way values require using your public methods (setEmail() and reset()).

silbermm: I would typically agree with this @samueljmello , but this is an ES6 class, not Typescript and ES6 does not support setting properties outside of methods, so the constructor is the only place to set them up. There are some tricks to make the properties 'private', but I think it overly complicates what we are trying to do here.

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
```diff
@@ -234,6 +244,14 @@
       vm.startAttendingOpen = true;
     }
 
+    function emailRequired() {
+      if (!editingOtherProfile() || (!emailChange.originalEmail)) {
+        return vm.requireEmail;
```
charlie-retzler: Is it confusing that we have a requireEmail property and an emailRequired with slightly different implementations.

crossroads.net/app/common/profile/personal/emailChange.service.js
```diff
@@ -0,0 +1,22 @@
+/* ngInject */
+class EmailChangeService {
```
charlie-retzler: Is this a singleton, and if so how do we make sure it gets called to clear everything out in each case before we start using it? I see several calls to reset on saving and in your family selection code, but what about if you go through your family selection tool, select another user, and then go to the profile. Does reset get run properly?

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
```diff
@@ -269,7 +287,13 @@
 
         if (vm.pform.email !== undefined) {
           if (vm.pform.email.$touched === true) {
-            vm.emailSet = true;
+            if (vm.forTrips) {
+              if (vm.oldEmail !== null && vm.oldEmail !== '') {
+                vm.emailSet = true;
```
IngageKrizJ: Have not followed the logic all the way through, but noticed the emailChange service has an isSet that isn't used anywhere outside the service - is this supposed to use that instead of vm.emailSet?

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
```diff
@@ -269,7 +287,13 @@
 
         if (vm.pform.email !== undefined) {
           if (vm.pform.email.$touched === true) {
-            vm.emailSet = true;
+            if (vm.forTrips) {
+              if (vm.oldEmail !== null && vm.oldEmail !== '') {
+                vm.emailSet = true;
+              }
+            } else {
+              vm.emailSet = true;
```
IngageKrizJ: Same as comment on [line 292](https://github.com/crdschurch/crds-angular/pull/2822/files#diff-5b0f97dbeabe3271ebb3b33e34c1884fR292)

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
charlie-retzler: Not a huge fan of all the chained if's, but not sure how to simplify.

andycanterbury: We want to make sure the email address only gets set once so we use the isSet flag to tell us if its ok to change it. vm.emailSet is used to flag if we need to show the password prompt when changing email or password.

crossroads.net/app/common/profile/personal/emailChange.service.js
charlie-retzler: I guess what I'm trying to say it would be good to call reset before we start using it rather than after we are done using it.

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
IngageKrizJ: Ah, got it, thanks.

crossroads.net/app/common/profile/personal/emailChange.service.js
silbermm: Good question and one we struggled with. The problem is that if we always call it before we begin using it, than it defeats the purpose. 
Example:

A person starts the trip application at the profile page. Profile page loads and resets the service. Person moves to next page and then back to profile page. We call reset again and wipes out the state that we were trying to preserve. 

But if we reset when they are done with the profile, things happen in the correct order. 

We are resetting when a new person is picked in the Family tool as a precaution. 

Make sense?

crossroads.net/app/common/profile/personal/profilePersonal.controller.js
```diff
@@ -136,12 +142,16 @@
           vm.viewReady = true;
 
         }
-
+        emailChange.setEmail(vm.profileData.person.emailAddress);
```
charlie-retzler: per conversation I think this belongs in the Profile.Personal.get and in the else below it since we may not have vm.profileData loaded yet.

PR #2854, opened 2016-09-14
DE1937 Fix Detailed Trip Report

silbermm:
The Stored Procedure had the wrong id's for both new attributes. This is
because I gave @plachmann bad information... Please forgive me @plachmann!

:cold_sweat:


plachmann:
@silbermm You are forgiven!

PR #2865, opened 2016-09-15
Defect/de1940 logged out when 

silbermm:


Gateway/crds-angular/Services/PersonService.cs
```diff
@@ -62,7 +62,8 @@ public void SetProfile(String token, Person person)
             // TODO: See About moving the check for new password above or moving the update for user / person into an atomic operation
             //
             // update the user values if the email and/or password has changed
-            if (!(String.IsNullOrEmpty(person.NewPassword)) || (person.EmailAddress != person.OldEmail))
+            //var user = _userService.
```
andycanterbury: Commented code here

andycanterbury: *you're

PR #2872, opened 2016-09-19
Feature/us4748 script fix

jpc11623:


IngageKrizJ:
Change looks good, but don't want to merge since Demo has already been deployed.  Should talk with Thundercats about this, since it will impact the prod deployment (either we merge this fix now, or we have manual steps after prod deploy).

CI/SQL/20160913_074800_US4748-CRDS-groups-report.sql
```diff
@@ -28,7 +28,7 @@ BEGIN
 			   (302
 			   ,N'CRDS Groups Snapshot Summary'
```
charlie-retzler: Report name should not have CRDS in it

charlie-retzler: We will need to treat this as a hotfix so we will need the PR created against Release branch.

jpc11623:
I'm hesitant to remove "CRDS" from the report name, as it's a duplicate of an existing MP report, and there is not an easy way at a glance to differentiate this from the stock MP report, other than the directory it's in. I can change this easily enough on int and demo, but wanted to bring this point up for further discussion, because I think it incurs some risk by not explicitly naming these as CRDS reports.

PR #2879, opened 2016-09-22
Feature/us4205 refactorparticipantmodel

lknair2015:


IngageKrizJ:
I reviewed, it looks good to me.  I also pulled this branch, ran locally, and tested things that could be impacted that I knew how to test: Registering new user, Community Groups signup & wait list, and Group Tool (various actions involving participants).  @IngageRitterS , @jpc11623 , and @silbermm, any objections to merging this?  I know there are other areas that are possibly impacted that I did not test, but looking at the changes, everything looks like it should work fine.  These models were not exposed to the frontend, they were only used between the Translation layer and crds-angular.

Gateway/MinistryPlatform.Translation/Models/MpParticipant.cs
```diff
@@ -12,5 +12,8 @@ public class Participant
         public int Age { get; set; }
         public DateTime? AttendanceStart { get; set; }
         public bool ApprovedSmallGroupLeader { get; set; }
+        public string role { get; set; }
```
silbermm: for consistency, can we update `role` to `Role`?

IngageKrizJ: Good catch.

lknair2015: Updated the field to Role

PR #2880, opened 2016-09-22
US4709 - Edit Group title should be a content block

tallented:
The title and description paragraph are now a content block matching the structure of the Create Group html page

edit group content block


jpc11623:
Reviewed, looks good.

PR #2883, opened 2016-09-23
Feature/us5101 rename formly types

JoeKer1:
Whoever reviews this: Should we just take the crds prefix off? And I removed the debugger line from create group service and set one of the properties of a field to true instead of 'true' to fix some error messages in the console.


crossroads.net/app/formlyBuilder/formlyConfig/types/crdsInput.js
```diff
@@ -1,9 +1,9 @@
 export default ngModule => {
-    ngModule.config(addFormlyBuilderCheckbox);
+    ngModule.config(addCrdsCheckbox);
 
-    function addFormlyBuilderCheckbox(formlyConfigProvider) {
```
Doug-Shannon: This was named wrong in the first place, should not be addCrdsCheckbox, but it should be addCRDSInput.  Also, I think in all of these, CRDS should be all caps or all lowercase instead of camelcase since it is an acronym.

Doug-Shannon: It is wrong in a few different places

PR #2924, opened 2016-10-07
Defect/de1965 group tool links

k8tied1:


Gateway/MinistryPlatform.Translation/Repositories/InvitationlRepository.cs
```diff
@@ -103,7 +103,7 @@ public void MarkInvitationAsUsed(string invitationGuid)
             
                 var dictionary = new Dictionary<string, object>();
                 dictionary.Add("Invitation_ID", invitation.InvitationId);
-                dictionary.Add("Invitation_GUID", invitationGuid);
+                dictionary.Add("Invitation_GUID", new Guid(invitationGuid));
```
jpc11623: I am curious about this conversion being necessary - doesn't sql server automatically cast a string to a guid if the column type is guid?

k8tied1: It probably should. I was getting a 'link expired' error when running locally without the forced conversion because it was failing that it wanted a GUID instead of a string.

jpc11623:
Reviewed, looks good.

PR #2939, opened 2016-10-11
Feature/us4966 email text reminder time in user timezone

vchtelmakh:
https://rally1.rallydev.com/#/66096127196d/detail/userstory/62678569135


crossroads.net/app/live_stream/streaming_reminder/streamingReminder.controller.js
```diff
@@ -113,7 +114,8 @@ export default class StreamingReminderController {
     this.model.day = day;
     
     if (day) {
-      this.model.day = day.start.format(this.dateFormats.key)
+      let event = new Event();
+      this.model.day = event.formatGeneralDateTimeToLocalDate(day.start);
     } else {
```
richardbarker: This is used as an instance level method.  Looks like it could be static.

PR #2953, opened 2016-10-12
Started Configuring Eslint

Ganonside:
I removed all of the previous rules from the .eslintrc file because they
were old.

The .eslintrc file now reads in airbnb rules.


crossroads.net/.eslintrc
```diff
@@ -1,55 +1,8 @@
...
-    "notypeof": false,
-    "proto": false,
-    "scripturl": false,
-    "shadow": false,
-    "sub": true,
-    "supernew": false,
-    "validthis": false,
-    "noyield": false,
+  "rules": {
+    "comma-dangle": 0
```
IngageKrizJ: According to [documentation on eslint.org](http://eslint.org/docs/rules/comma-dangle#options), this has values like never and always.  Does 0 work as well?

Ganonside: 0 - ignore
1 - warn
2 - error

IngageKrizJ: Ok, we're ignoring (allowing) dangling commas.  Thanks for the info.

PR #2969, opened 2016-10-13
Feature/us4998 my camps dashboard

silbermm:


Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -498,6 +498,7 @@
   <add key="SummerCampForm.PreferredRoommate" value="1505"/>
   <add key="CreateContactStoredProc" value="api_crds_CreateContact"/>
   <add key="CampEventType" value="8" />
+  <add key="CampEventTypeName" value="Camp"/>
```
silbermm: why use the name of the camp and not the id?

Gateway/crds-angular/Services/CampService.cs
```diff
@@ -95,5 +100,43 @@ public void SaveCampReservation(CampReservationDTO campReservation, int eventId,
             
             _formSubmissionRepository.SubmitFormResponse(formResponse);
         }
+
+        public List<MyCampDTO> GetMyCampInfo(string token)
```
silbermm: this is a fairly large method and should probably have some tests

PR #2976, opened 2016-10-14
DE2026: Changed default user count to 1, verbage to "person" & "is",  and updated tests

samueljmello:


richardbarker:
Looks good.

PR #3003, opened 2016-10-17
Defects DE2036 & DE2047

vchtelmakh:
Fixed both defects on one PR because they are related and I needed to make sure all the tests pass regardless. 

https://rally1.rallydev.com/#/66096127196d/detail/defect/69082782852
https://rally1.rallydev.com/#/66096127196d/detail/defect/69395005952


crossroads.net/app/live_stream/services/stream-status.service.js
```diff
@@ -1,16 +1,66 @@
...
+    var deferred = this.q.defer();
+
+    let url = `${this.url}broadcaster/${this.ssid}/events`;
+
+    return this.resource(url, {}, { get: { method: 'GET', headers: this.headers } })
+        .get()
+        .$promise
+        .then((response) => {
+          let events = response.data.events;
+          events = this.parseEventsWithParam(events);
```
mjvcallibrity: Is there a better way of doing this? Is seems strange to assign `events` to `response.data.events` and then on the next line reassign `events` to the result of the `parseEventsWithParam()` call.

crossroads.net/app/live_stream/services/stream-status.service.js
```diff
@@ -1,16 +1,66 @@
...
+          let isBroadcasting = this.isBroadcasting(events);
+          let streamStatus = this.determineStreamStatus(events, isBroadcasting);
+          this.streamStatus = streamStatus;
```
mjvcallibrity: Would it be good practice to reduce this to one line? e.g.

`this.streamStatus = this.determineStreamStatus(events, isBroadcasting);`

crossroads.net/app/live_stream/services/stream-status.service.js
```diff
@@ -1,16 +1,66 @@
...
+          deferred.resolve(events);
+        });
+
+    return deferred;
+
+  };
+
+  parseEventsWithParam(events) {
```
mjvcallibrity: The parameter for this function is an array of events, but the name of the function makes it sound like the parameter should be a string of the param name with which you are parsing. Could use a rename.

crossroads.net/app/live_stream/services/stream-status.service.js
```diff
@@ -81,5 +131,30 @@ export default class StreamStatusService {
     return isEventStartBeforeCurrentTime;
   }
 
+  isBroadcasting(events){
+    let areAnyEventsBroadcasting = false;
+
+    for(let idx=0; idx<events.length; idx++){
```
mjvcallibrity: For the sake of consistency with the rest of the codebase, I would change `idx` to just `i` throughout this `for` loop.

crossroads.net/app/live_stream/services/stream-status.service.js
```diff
@@ -81,5 +131,30 @@ export default class StreamStatusService {
...
+    return areAnyEventsBroadcasting;
+  };
+
+  isEventCurrentlyLive(event){
+    let currentTime = moment();
+    let eventStartTime = moment(event.start);
+    let eventEndTime = moment(event.end);
+    let isEventStartBeforeCurrentTime = eventStartTime.isBefore(currentTime);
+    let isEventEndAfterCurrentTime = eventEndTime.isAfter(currentTime);
+    let isEventLive = isEventStartBeforeCurrentTime && isEventEndAfterCurrentTime;
```
mjvcallibrity: I think that the momentjs methods `isBefore()` and `isAfter()` are descriptive enough that you could probably get rid of your `isEventStartBeforeCurrentTime` and `isEventEndAfterCurrentTime` variable declarations and instead go with something like:

let isEventLive = eventStartTime.isBefore(currentTime) && eventEndTime.isAfter(currentTime);

return isEventLive

crossroads.net/spec-es6/live_streaming/services/stream-status.service.spec.js
```diff
@@ -39,38 +41,106 @@ describe('Stream Status Service', () => {
     }
   ];
 
+  let eventsWithLiveEvent = [
+    {
+      "start": moment(baseTime).subtract(4, 'hour').format('YYYY-MM-DD H:mm:ss'),
```
mjvcallibrity: Super nitpicky, I know, but momentjs is able to handle time strings both in the singular and plural format. So, for example, `.subtract(4, 'hour')` could instead be `.subtract(4, 'hours')`. I've seen this done in other parts of the codebase as well, so for sake of style consistency, it might be best to go ahead and pluralize these here.

crossroads.net/spec-es6/live_streaming/services/stream-status.service.spec.js
```diff
@@ -39,38 +41,106 @@ describe('Stream Status Service', () => {
...
+    it('should indicate that the event starts after the current time', () => {
+      eventStartingAfterCurrentTime = events[2];
+      doesEventStartAfterCurrentTime = streamStatusService.doesEventStartAfterCurrentTime(eventStartingAfterCurrentTime);
+      expect(doesEventStartAfterCurrentTime).toBeTruthy();
+    });
+
+    it('get hours until next streaming event starts', () => {
+      hrsToNextEvent = streamStatusService.getHoursToNextEvent(events);
+      expectedHrsToEvent = 1;
+      expect(1).toEqual(expectedHrsToEvent);
```
mjvcallibrity: This appears to be checking that 1 equals 1, instead of checking that the result of a method call equals an expected result. This should be fixed.

crossroads.net/spec-es6/live_streaming/services/stream-status.service.spec.js
```diff
@@ -39,38 +41,106 @@ describe('Stream Status Service', () => {
...
+      isBroadcasting = false;
+      streamStatus = streamStatusService.determineStreamStatus(events, isBroadcasting);
+      upcomingStatus = CONSTANTS.STREAM_STATUS.UPCOMING;
+      expect(streamStatus).toEqual(upcomingStatus);
+    });
+
+    it('should indicate that stream status changed', () => {
+      isBroadcasting = false;
+      didStreamStatusChange = streamStatusService.didStreamStatusChange(events, isBroadcasting);
+      expect(streamStatus).toBeTruthy(didStreamStatusChange);
```
mjvcallibrity: I am not familiar with instances in which `toBeTruthy()` should be passed an argument. I was under the impression that `toBeTruthy()` was a method that did not accept parameters. I'm not sure that this test is functioning as expected.

crossroads.net/app/live_stream/services/stream-status.service.js
vchtelmakh: Yea, that local streamStatus var was definitely redundant.

PR #3016, opened 2016-10-20
US5327 - throw it against the wall; see if it sticks

nogorilla:


beckysingson:
@nogorilla I can't really test this on an actual device but it looks pretty good in Browserstack on iOS.

PR #3018, opened 2016-10-20
Feature/us5078 fetch quick amounts backend service

vchtelmakh:
https://rally1.rallydev.com/#/41662702253d/detail/userstory/63442555006


richardbarker:
My only suggestion would be to refactor the DTO and method names to reflect the new table name, i.e. 
GetPredefinedDonationAmounts vs GetQuickDonationAmounts

PR #3023, opened 2016-10-21
DE2053 - move success flag

nogorilla:
/cc @beckysingson


beckysingson:
@nogorilla This worked. The timeout timing was fine. However, when I submit and then after the modal goes away, I get taken back to /live instead of the stream page (even though a service was live). Not sure if related to this or #3029

nogorilla:
@beckysingson couldn't reproduce the error you mention. Can you retest?

beckysingson:
@nogorilla fluke? fine second time around.

PR #3027, opened 2016-10-21
Feature/us5316 camp family select

silbermm:
Create the family select page at `/camps/:campId/family`


Gateway/MinistryPlatform.Translation/Models/MpStoredProcBool.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace MinistryPlatform.Translation.Models
+{
+    public class MpStoredProcBool
```
andycanterbury: There has to be a better way to get a single value from the new rest api... 🙁

crossroads.net/.eslintrc
```diff
@@ -26,6 +26,7 @@
     "crds_utilities" : false,
     "spyOn": false,
     "_$location_" : false,
-    "expect" : false
+    "expect" : false,
+    __API_ENDPOINT__: false
```
andycanterbury: ??

crossroads.net/spec/camps/camps.service.spec.js
```diff
@@ -36,8 +37,13 @@ describe('Camp Service', () => {
     httpBackend.flush();
   });
 
+  it('should make the API call to get my summer camp family', () => {
```
andycanterbury: Does this test not test anything?

crossroads.net/spec/camps/camps.service.spec.js
```diff
@@ -36,8 +37,13 @@ describe('Camp Service', () => {
...
 
+  });
+
+  it('should make the API call to get my summer camp family and handle error', () => {
```
andycanterbury: And this one?

PR #3036, opened 2016-10-25
Defect/de2061 use imgix for streaming images

richardbarker:


crossroads.net/app/live_stream/current_series/currentSeries.controller.js
```diff
@@ -1,8 +1,8 @@
 
 export default class CurrentSeriesController {
-  constructor(CMSService, $modal) {
+  constructor(CMSService, $modal, ResponsiveImageService) {
     this.title        = '';
-    this.picture      = '';
+    this.picture      = undefined;
```
samueljmello: You can also just do:

`this.picture;`

PR #3042, opened 2016-10-26
Feature/us5096 get previous gift amount

sseissiger:


Gateway/MinistryPlatform.Translation.Test/Services/DonorServiceTest.cs
```diff
@@ -1329,6 +1336,126 @@ public void TestGetDonationsForAuthenticatedUser()
...
+                    {"Donation_Status", "Succeeded"}
+                },
+                new Dictionary<string, object>
+                {
+                    {"dp_RecordID", 4},
+                    {"Display_On_Giving_History", false},
+                    {"Display_On_Statements", false},
+                    {"Display_On_MyTrips", false},
+                    {"Donation_Status", "Succeeded"}
+                }
```
samueljmello: Just curious; but for these redundant dictionary objects, why not just create one and then modify the changed values before adding to the main dictionary array? Seems like a lot of extra lines.

sseissiger: I followed the rest of the tests in the file.  Yes seems like a lot of lines, but easy to see the full objects that are available.

PR #3070, opened 2016-10-27
DE2114 - geolocation feedback

nogorilla:


beckysingson:
@nogorilla Can you make it completely null instead of "0 people"? Like underlined & blank more like the default state of zip code?

beckysingson:
@nogorilla I'm so sorry, I misstated the intention on the update to the Dismiss button - the new look for the button should apply on both desktop and mobile (my specification about "on mobile" was just that people were most struggling with it's placement on a mobile device but they want the style change on that across devices).

PR #3076, opened 2016-10-27
Feature/us5183 leader sends individuals messages

Doug-Shannon:
fixing defect for enddate not considering time


Gateway/MinistryPlatform.Translation.Test/Services/GroupParticipantRepositoryTest.cs
```diff
@@ -114,8 +114,8 @@ public void GetAllGroupNamesLeadByParticipantShouldTransfomGroupParticipantIntoG
 
             string search = $"Group_Participants.participant_id = {participantId}" +
                             $" AND Group_Role_ID = {GroupLeaderRole}" +
-                            $" AND (Group_ID_Table.End_Date > '{DateTime.Today:yyyy-MM-dd}' OR Group_ID_Table.End_Date Is Null)" +
```
JoeKer1: Can we just do AND (Group_ID_Table.End_Date > GetDate() OR ....

It works in the RestApi, not sure which is better.

PR #3111, opened 2016-11-03
US5229 Api Versioning

eymiha:
Pulling the library code into the project instead of putting it into a NuGet package right now. (No extra changes required to include a package vs included code.)

See Crossroads.ApiVersioning/readme.txt for usage details. 1.0.0 Versioned routes have been added for all existing routes, and the existing routes have been retained as versionless routes. The new routes have also been made consistent (camelCased URL parts and argument names) and have been verified in Swagger.

One route in the GroupToolController has been *shrouded* from Swagger because it is affected by a known error in the Swagger page generation.


Gateway/crds-angular/Controllers/API/ChildcareController.cs
```diff
@@ -92,8 +95,9 @@ public IHttpActionResult ChildcareEventById(int eventid)
 
 
         [ResponseType(typeof(List<FamilyMember>))]
-        [Route("api/childcare/eligible-children")]
-        [AcceptVerbs("GET")]
+        [VersionedRoute(template: "childcare/eligableChildren", minimumVersion: "1.0.0")]
```
IngageKrizJ: misspelt - childcare/eligibleChildren

Gateway/crds-angular/Controllers/API/CongregationController.cs
```diff
@@ -20,20 +21,21 @@ public CongregationController(ICongregationService congregationService)
         }
 
         [ResponseType(typeof (Congregation))]
-        [Route("api/congregation/{id}")]
-        public IHttpActionResult Get(int id)
+        [VersionedRoute(template: "congregation/{congregationId}", minimumVersion: "1.0.0")]
+        [Route("congregation/{id}")]
+        public IHttpActionResult Get(int congregationId)
```
IngageKrizJ: Do you know if this is still working?  I thought by default that the parameter name in the route ({id}) had to match the parameter name on the method signature (int congregationId).  If they do not match, I thought you had to use a [FromUri(name="id")] to make sure it gets mapped from the route param to the method.

Gateway/crds-angular/Controllers/API/CongregationController.cs
```diff
@@ -42,19 +44,20 @@ public IHttpActionResult Get(int id)
         }
 
         [ResponseType(typeof (List<Room>))]
-        [Route("api/congregation/{id}/rooms")]
-        public IHttpActionResult GetRooms(int id)
+        [VersionedRoute(template: "congregation/{congregationId}/rooms", minimumVersion: "1.0.0")]
+        [Route("congregation/{id}/rooms")]
+        public IHttpActionResult GetRooms(int congregationId)
```
IngageKrizJ: Same as above.

Gateway/crds-angular/Controllers/API/CongregationController.cs
```diff
@@ -63,19 +66,20 @@ public IHttpActionResult GetRooms(int id)
         }
 
         [ResponseType(typeof (List<RoomEquipment>))]
-        [Route("api/congregation/{id}/equipment")]
-        public IHttpActionResult GetEquipment(int id)
+        [VersionedRoute(template: "congregation/{congregationId}/equipment", minimumVersion: "1.0.0")]
+        [Route("congregation/{id}/equipment")]
+        public IHttpActionResult GetEquipment(int congregationId)
```
IngageKrizJ: Same as above.

Gateway/crds-angular/Controllers/API/DonorStatementController.cs
```diff
@@ -39,7 +41,8 @@ public IHttpActionResult Get()
             }));
         }
 
-        [Route("api/donor-statement")]
+        [VersionedRoute(template: "donor-statement", minimumVersion: "1.0.0")]
```
IngageKrizJ: Should this be donorStatement to be consistent with above route?

Gateway/crds-angular/Controllers/API/GoVolunteerController.cs
```diff
@@ -143,14 +149,15 @@ public IHttpActionResult GetGetGroupConnectorsOpenOrgs(int initiativeId)
...
         }
 
-        [HttpGet]
         [ResponseType(typeof (List<GroupConnector>))]
-        [Route("api/group-connectors/{orgId}/{initiativeId}")]
-        public IHttpActionResult GetGroupConnectorsForOrg(int orgId, int initiativeId)
+        [VersionedRoute(template: "groupConnectors/{organizationId}/{initiativeId}", minimumVersion: "1.0.0")]
+        [Route("group-connectors/{orgId}/{initiativeId}")]
+        [HttpGet]
+        public IHttpActionResult GetGroupConnectorsForOrg(int organizationId, int initiativeId)
```
IngageKrizJ: Similar to above comment - are we certain this works with diff param names in the route versus in the method signature?

Gateway/crds-angular/Controllers/API/GroupToolController.cs
```diff
@@ -122,7 +125,8 @@ public IHttpActionResult EndSmallGroup([FromUri]int groupId, [FromUri]int groupR
         /// <param name="removalMessage">An optional message to send to the participant when they are removed.  This is sent along with a boilerplate message.</param>
         /// <returns>An empty response with 200 status code if everything worked, 403 if the caller does not have permission to remove a participant, or another non-success status code on any other failure</returns>
         [RequiresAuthorization]
-        [Route("api/grouptool/grouptype/{groupTypeId:int}/group/{groupId:int}/participant/{groupParticipantId:int}")]
+        [VersionedRoute(template: "grouptool/groupType/{groupTypeId}/group/{groupId}/participant/{groupParticipantId}", minimumVersion: "1.0.0")]
```
IngageKrizJ: Any reason for removing the int specifier on the groupTypeId, groupId, and groupParticipantId route parameters?  It is not required, but it is a check to make sure the route has "valid" data when it comes in (well, the route would never get invoked if you sent "abc" for groupId, for instance).

Gateway/crds-angular/Crossroads.ApiVersioning/ApiRouteProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+                VersionConstraint constraint = null;
+                if (routeEntry.Route.Constraints.ContainsKey("allowedVersions"))
+                    constraint = routeEntry.Route.Constraints["allowedVersions"] as VersionConstraint;
+                VersionSpace.Add(route, constraint);
+            }
+            int routeCount = VersionSpace.Count();
+            JArray problems = VersionSpace.Problems();
+            if (problems.Count > 0)
+            {
+                System.Diagnostics.Debug.WriteLine("----------------------- Controller: " + controllerDescriptor.ControllerName);
```
IngageKrizJ: Could we use log4net here instead of System.Diagnostics?  I'm not familiar with System.Diagnostics, so not sure where this output would show up.

eymiha: System.Diagnostics.Debug only shows up when running in debug. I figured if new routes are being written, they'd be tested in debug. If that's a bad assumption we can switch to log4net

Gateway/crds-angular/Controllers/API/GroupToolController.cs
eymiha: We were applying it sporadically. If we're going to do it, we ought to do it everywhere. But if we're using Swagger, the types are documented.

Gateway/crds-angular/Controllers/API/CongregationController.cs
eymiha: Fixed.

Gateway/crds-angular/Controllers/API/CongregationController.cs
eymiha: Fixed.

Gateway/crds-angular/Controllers/API/DonorStatementController.cs
eymiha: Good catch. It got by me.

Fixed.

Gateway/crds-angular/Controllers/API/GoVolunteerController.cs
eymiha: Fixed, by using the longer name in the old api call as well as the new.

Gateway/crds-angular/Controllers/API/CongregationController.cs
eymiha: I'm curious though that some of our old calls (such as in the CampController) specified eventid as the parameter and eventId as the method argument. There're strange things 
afoot.

eymiha: I'm passing through all the controller to sync everything up.

Gateway/crds-angular/Crossroads.ApiVersioning/ApiRouteProvider.cs
eymiha: Now using log4net to always emit versioning integrity errors

PR #3112, opened 2016-11-03
Feature/us5308 sso configure domain cookie

richardbarker:


crossroads.net/environment.config.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   get: function() {
     return {
+      __CRDS_ENV__: JSON.stringify(process.env.CRDS_ENV),
```
samueljmello: Is there a reason why we don't want to provide defaults here? Is it so localhost doesn't get borked?

PR #3115, opened 2016-11-03
DE1980: Added KC Group Sort column with default of 999 and updated We…

k8tied1:
…ekend

Services Report to use that for sorting instead of attribute

### DIFF
--- a/CI/SQL/20161102_134920_DE1980-AddPageViewForGroupSortingForWeekendServiceReport.sql
+++ b/CI/SQL/20161102_134920_DE1980-AddPageViewForGroupSortingForWeekendServiceReport.sql
@@ -1,0 +1,20 @@
+USE [MinistryPlatform]
+GO
+
+INSERT INTO [dbo].[dp_Page_Views]
+           ([View_Title]
+           ,[Page_ID]
+           ,[Description]
+           ,[Field_List]
+           ,[View_Clause]
+           ,[Order_By])
+     VALUES
+           ('KC Group Sort'
+           ,322 --Groups
+           ,'Set sort order for Weekend Service Report'
+           ,'Groups.KC_Sort_Order, Groups.Group_Name'
+           ,'Ministry_ID_Table.[Ministry_Name] = ''Kids Club'' AND Groups.[End_Date] IS NULL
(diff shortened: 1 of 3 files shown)


CI/SQL/20161102_141100_DE1980-UpdateSP_report_weekend_service_Crossroads_ForGroupSortingForWeekendServiceReport.sql
```diff
@@ -0,0 +1,255 @@
...
+    , O.Shift_End
+    , O.Room
+    , G.Group_ID
+    , Con.Congregation_Name
+	, R.Response_ID
+	, R.Response_Date
+	, RowNum = ROW_NUMBER() OVER (PARTITION BY O.Opportunity_ID, E.Event_ID ORDER BY E.Event_ID,R.Response_Date)
+	, Position = 'MAX'
+	, DisplayColumn = CAST(0 AS INT)
+	, ISNULL(G.Group_Sort_Order,999) AS Sort_Order
```
k8tied1: That is correct. My concern is that if someone decides to remove the non-null requirement they would have to know to go update the stored proc, which is unlikely. I like the additional safety check :-)

CI/SQL/20161102_134920_DE1980-AddPageViewForGroupSortingForWeekendServiceReport.sql
```diff
@@ -0,0 +1,20 @@
...
+           ,[Field_List]
+           ,[View_Clause]
+           ,[Order_By])
+     VALUES
+           ('KC Group Sort'
+           ,322 --Groups
+           ,'Set sort order for Weekend Service Report'
+           ,'Groups.KC_Sort_Order, Groups.Group_Name'
+           ,'Ministry_ID_Table.[Ministry_Name] = ''Kids Club'' AND Groups.[End_Date] IS NULL
+			 AND (Groups.End_Date IS NULL OR Groups.End_Date >= GetDate())'
```
k8tied1: I really really didn't want to have groups where the end date is null.

PR #3126, opened 2016-11-07
Defect/de2147 error message saving start date

k8tied1:
Notes - I went ahead and pulled the hotfix into my branch because I was hoping that it might fix my issues. It didn't but I figured I'd leave it instead of ripping it back out.


CI/SQL/20161107_114600_DE2147-ErrorOnStartDate.sql
```diff
@@ -0,0 +1,14 @@
+USE [MinistryPlatform]
+GO
+
+UPDATE dp_page_views 
+SET Field_List = 'Contact_ID_Table.[Contact_ID] AS [Contact ID] 
+    , Contact_ID_Table.[Display_Name] AS [Display Name]
+    , Contact_ID_Table.[Nickname] AS [Nickname]
+    , Contact_ID_Table.[First_Name] AS [First Name]
+    , Contact_ID_Table.[Email_Address] AS [Email Address]
+    , Participants.[Participant_ID] AS [Participant ID]
```
IngageRitterS: Can we either remove the AS statements or change the field names to have underscores?   I realize this is old code, but this change would follow our standards.

Gateway/crds-angular/Services/PersonService.cs
```diff
@@ -47,13 +46,11 @@ public void SetProfile(String token, Person person)
...
             _objectAttributeService.SaveObjectAttributes(person.ContactId, person.AttributeTypes, person.SingleAttributes, configuration);
 
-            Participant participant = _participantService.GetParticipant(person.ContactId);
+            var participant = _participantService.GetParticipant(person.ContactId);
             if (participant.AttendanceStart != person.AttendanceStartDate)
             {                
                 participant.AttendanceStart = person.AttendanceStartDate;
-                // convert to the object with underscores
-                var p = Mapper.Map <Participant>(participant);
-                _participantService.UpdateParticipant(getDictionary(p));
```
IngageRitterS: Is this mapping used anywhere else?  If not can we remove crds_angular.Models.Crossroads.Participants?

k8tied1: There is no crds_angular.Models.Crossroads.Participants. Participant was MpParticipant. I removed the using statement on the top and switched this to var.

CI/SQL/20161107_114600_DE2147-ErrorOnStartDate.sql
k8tied1: yep

PR #3127, opened 2016-11-08
US5228-ProductSummaryPage

plachmann:


Gateway/crds-angular.test/Services/CampServiceTest.cs
```diff
@@ -498,6 +504,62 @@ public void shouldGetCamperInfoNoGrade()
...
+                ProductId = 111,
+                BasePrice = 1000,
+                DepositPrice = 200,
+                ProductName = "Hipster Beard Wax"
+            };
+
+            var mpevent = new MpEvent
+            {
+                EventId = 999,
+                EventTitle = "Hipster Beard Training",
```
andycanterbury: 😆

Gateway/crds-angular/Services/CampService.cs
```diff
@@ -323,6 +347,43 @@ public void SaveWaivers(string token, int eventId, int contactId, List<CampWaive
             _eventRepository.SetWaivers(waiverResponses);
         }
 
+        public void SaveInvoice(CampProductDTO campProductDto, string token)
+        {
+            // set finainacial assistance flag in form response
+            var participant = _participantRepository.GetParticipant(campProductDto.ContactId);
+            var eventParticipantId = _eventRepository.SafeRegisterParticipant(campProductDto.EventId, participant.ParticipantId);
```
andycanterbury: We are making them a participant here? I suspect this is after the waivers, and we need them to already be a participant by then.

crossroads.net/app/camps/application_page/product_summary/product_summary.controller.js
```diff
@@ -0,0 +1,74 @@
...
+        option.isCurrent = true;
+      } else {
+        this.isBasePriceCurrent = true;
+      }
+    }
+  }
+
+  getMinimumDeposit() {
+    if (this.model.financialAssistance) {
+      return 50;
```
andycanterbury: I know that 50 is kind of digital stone, but should we at least put it in a variable in case they ever do want to change it?

crossroads.net/app/camps/camps.routes.js
```diff
@@ -46,6 +51,9 @@ export default function CampRoutes($stateProvider) {
         campsService: 'CampsService',
         getCampInfo,
         $stateParams: '$stateParams',
+
+        // FIXME: do we really need to resolve family for every page that lives under `campsignup`?
```
andycanterbury: Do we need to fix this "FIXME"?

Gateway/crds-angular/Services/CampService.cs
plachmann: Now using GetEventParticipantRecordId

crossroads.net/app/camps/application_page/product_summary/product_summary.controller.js
plachmann: this is still digital stone but now much prettier in a class variable

Gateway/crds-angular/Services/CampService.cs
```diff
@@ -323,6 +347,43 @@ public void SaveWaivers(string token, int eventId, int contactId, List<CampWaive
             _eventRepository.SetWaivers(waiverResponses);
         }
 
+        public void SaveInvoice(CampProductDTO campProductDto, string token)
+        {
+            // set finainacial assistance flag in form response
```
plachmann: Agree. Fixed

PR #3131, opened 2016-11-09
DE2184-Use campers contact id on call to GetCampProductDetails

plachmann:


crossroads.net/app/camps/camps.service.js
```diff
@@ -17,7 +17,7 @@ class CampsService {
     // eslint-disable-next-line prefer-template
     this.campWaiversResource = $resource(__API_ENDPOINT__ + 'api/camps/:campId/waivers/:contactId', { campId: '@campId', contactId: '@contactId' });
     // eslint-disable-next-line prefer-template
-    this.productSummaryResource = $resource(__API_ENDPOINT__ + 'api/camps/:campId/product', { campId: '@campId' });
+    this.productSummaryResource = $resource(__API_ENDPOINT__ + 'api/camps/:campId/product/:camperId', { campId: '@campId', camperId: '@camperId' });
```
Ganonside: Is camperId different from contactId on purpose. If it's meant to be different then I think that's ok but everywhere else it seems to be contactId.

Gateway/MinistryPlatform.Translation/Repositories/FormSubmissionRepository.cs
```diff
@@ -168,7 +168,7 @@ public string GetFormResponseAnswer(int formId, int contactId, int formFieldId)
             var formResponseAnswerId = GetFormResponseAnswerIdForFormFeildFormResponse(formResponseId,formFieldId);
             var responseAnswer =_ministryPlatformRestRepository.UsingAuthenticationToken(apiToken).Get<MpFormAnswer>(formResponseAnswerId, selectColumns);
 
-            return responseAnswer.Response;
+            return responseAnswer == null ? string.Empty:responseAnswer.Response;
```
Ganonside: I would add spaces around the ':' to improve readability.

crossroads.net/app/camps/camps.service.js
plachmann: intentional for clarity that this is the camper and not the logged in user

Gateway/MinistryPlatform.Translation/Repositories/FormSubmissionRepository.cs
plachmann: change made

PR #3151, opened 2016-11-11
Feature/us5378 2 submit payment

sseissiger:


richardbarker:
This change is adding an output to an API.  I wonder if this qualifies as a change that requires versioning?

sseissiger:
I am only adding an additional piece of data to the return object.  It would not be a breaking change, so I don't think it needs to be versioned

PR #3155, opened 2016-11-11
Change camelCase to lowercase-with-dashes in API non-parameter URI parts

eymiha:
As requested in the developer chapter meeting and API community or practice.


Gateway/crds-angular/Controllers/API/CampController.cs
```diff
@@ -42,7 +42,7 @@ public IHttpActionResult GetCampFamily(int eventId)
         }
 
         [ResponseType(typeof(List<MyCampDTO>))]
-        [VersionedRoute(template: "my-camp", minimumVersion: "1.0.0")]
+        [VersionedRoute(template: "myCamp", minimumVersion: "1.0.0")]
```
silbermm: this should be `my-camp`

Gateway/crds-angular/Controllers/API/CampController.cs
```diff
@@ -305,7 +305,27 @@ public IHttpActionResult SaveCamperEmergencyContact([FromBody] List<CampEmergenc
...
+                }
+                catch (Exception e)
+                {
+                    var apiError = new ApiErrorDto("Camp Confirmation failed", e);
+                    throw new HttpResponseException(apiError.HttpResponseMessage);
+                }
+            });
+        }
+
+        [VersionedRoute(template: "camps/{eventId}/emergencyContact/{contactId}", minimumVersion: "1.0.0")]
```
silbermm: should be `emergencycontact` (not camelCased)

eymiha:
I'd mistakenly replaced the whole camp controller instead of only the conflicted region. Should be good to go now.

PR #3160, opened 2016-11-13
Defect/de2214 deleteemergencycontact

lknair2015:


Gateway/crds-angular/Services/CampService.cs
```diff
@@ -154,6 +154,22 @@ public void SaveCamperEmergencyContactInfo(List<CampEmergencyContactDTO> emergen
             var participant = _participantRepository.GetParticipant(contactId);
             var eventParticipantId = _eventRepository.SafeRegisterParticipant(eventId, participant.ParticipantId);
             var answers = new List<MpFormAnswer>();
+            if (emergencyContacts.Count == 1)
```
silbermm: i'm having a hard time understanding how this will fix the defect. Won't this just add a new blank emergencyContact?

PR #3173, opened 2016-11-16
US5735 Payment Subpage on Batches

silbermm:
Add scripts for payments subpage on batch

payment_subpage


CI/SQL/20161116_101224_US5735-add-payments-subpage-to-batch.sql
```diff
@@ -0,0 +1,53 @@
+use MinistryPlatform
```
andycanterbury: USE not use. 😝

PR #3178, opened 2016-11-16
Defect/de2231 invalid form error

Ganonside:
Some eslint fixes and changing the error message for invalid forms

### DIFF
--- a/crossroads.net/app/camps/application_page/camper_info/camper_info.controller.js
+++ b/crossroads.net/app/camps/application_page/camper_info/camper_info.controller.js
@@ -17,6 +17,7 @@ class CamperInfoController {
 
   submit() {
     this.submitting = true;
+
     if (this.infoForm.$valid) {
       this.camperInfoForm.save(this.stateParams.campId).then(() => {
         this.rootScope.$emit('notify', this.rootScope.MESSAGES.successfullRegistration);
@@ -29,7 +30,7 @@ class CamperInfoController {
       });
     } else {
       this.submitting = false;
-      this.rootScope.$emit('notify', this.rootScope.MESSAGES.generalError);
+      this.rootScope.$emit('notify', this.rootScope.MESSAGES.invalidFormFields);
     }
   }
 }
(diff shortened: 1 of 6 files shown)


crossroads.net/spec/camps/camper_info/camper_info.component.spec.js
```diff
@@ -92,6 +92,7 @@ describe('Camper Info Component', () => {
   });
 
   it('should get new data when selecting another family member', () => {
-
-  })
+    // FIXME: do we need to implement this test case?
+    //   Jonathan Horner
+  });
```
Ganonside: I'm assuming this hasn't been implemented this because we aren't offering the selection of other family members after completion yet but i figured i should make sure.

Ganonside:
In the picture on the defect card there isn't any text underneath the field notifying the user that the field needs to be filled in. When I ran it on my machine that text was showing with no issue. I'm pretty sure I was on the right page (camper-info) but if that wasn't the right one please let me know.

crossroads.net/app/camps/application_page/camper_info/camper_info.controller.js
```diff
@@ -29,7 +30,7 @@ class CamperInfoController {
       });
     } else {
       this.submitting = false;
-      this.rootScope.$emit('notify', this.rootScope.MESSAGES.generalError);
+      this.rootScope.$emit('notify', this.rootScope.MESSAGES.invalidFormFields);
```
silbermm: is this a content block that you added to the CMS? If so, you'll need to script it ( see https://github.com/crdschurch/SS-CMS/tree/development/CI/SQL for examples that you can copy ) and create a pull request in SS-CMS repo. 

I would also ask if it's necessary that we use a different message here. We consistently use the `generalError` everywhere else.

andycanterbury: Does this need to be changed back now that the problem is solved?

PR #3205, opened 2016-11-28
Add an ID to the remove button the confirm removal dialog.

candrews-thrive:
The dialog is access from the giving tab on My Profile page.


simeonsena:
Looks good

PR #3220, opened 2016-12-01
Feature/us5915 kids club room roster

IngageKrizJ:
* [US5915 - Create Kids Club Room Roster](https://rally1.rallydev.com/#/27593764268/detail/userstory/78478532036)
* Decided to implement an SSRS report (with embedded dataset) rather than a page and/or page view
  * This gives the user a more straightforward way to get the roster, and it looks a lot better than what would come out of a page view export
  * Effort was comparable to a page and/or page view...
  * Sample report (exported as PDF): [Kids Club Event - Room Roster.pdf](https://github.com/crdschurch/crds-angular/files/625650/Kids.Club.Event.-.Room.Roster.pdf)


IngageKrizJ:
RecordID isn't working - need to look into this

PR #3227, opened 2016-12-02
US5235 Fix the borked build

silbermm:


CI/SQL/20161201_0934_US5235-add-program-to-product.sql
```diff
@@ -0,0 +1,12 @@
+use MinistryPlatform;
```
andycanterbury: Capital USE :-P

PR #3240, opened 2016-12-05
DE2305:Script to delete other households that are already primaries

abartoszek:


abartoszek:
Thanks Charlie!  ^_^



Adam Bartoszek
Crossroads Church
IT-Data Architect and Developer
513-225-2407

PR #3267, opened 2016-12-08
DE2396 Inline Giving Iframe Content Block

tcmacdonald:
This PR refactors the implementation of the inline giving iframe, per @bzmillerboy's feedback in DE2396. 

The content block referenced here should be populated across environments via the script included in the following PR...  

crdschurch/SS-CMS#203


crossroads.net/app/live_stream/stream/stream.controller.js
```diff
@@ -61,47 +58,30 @@ export default class StreamingController {
...
       interval: -16
-    }, ".donation-widget");
+    }, el);
   }
 
   buildUrl() {
-    return this.sce.trustAsResourceUrl(`${this.baseUrl}${this.queryStringParams}`);
+    const params = this.queryStringParams || 'type=donation&theme=dark';
+    return this.sce.trustAsResourceUrl(`${this.baseUrl}?${params}`);
   }
```
richardbarker: We could put a unit test around this.   Dan suggested anytime we have a defect we should make a test to detect it in the future.

PR #3268, opened 2016-12-08
Defect/de2273 back button

Ganonside:


crossroads.net/app/camps/application_page/camps_payment/camps_payment.controller.js
```diff
@@ -29,6 +20,10 @@ export default class CampPaymentController {
 
     // eslint-disable-next-line no-undef
     switch (__CRDS_ENV__) {
+      case 'local':
+        this.baseUrl = 'http://local.crossroads.net:8080';
```
andycanterbury: local.crossroads.net? Should this be localhost?

crossroads.net/app/camps/application_page/camps_payment/camps_payment.controller.js
```diff
@@ -29,6 +20,10 @@ export default class CampPaymentController {
...
+        this.returnUrl = 'http://local.crossroads.net:3000/camps';
```
andycanterbury: localhost?

crossroads.net/app/camps/camps.service.js
```diff
@@ -159,11 +165,13 @@ class CampsService {
   }
 
   getShirtSizes() {
+    /* eslint-disable new-cap */
```
andycanterbury: What is this linting rule we are ignoring? Is it a rule we should talk about getting rid of?

Ganonside: it just says that unless a function is a constructor it shouldn't start with a capital letter. We can't control that in this situation so I ignored it.

crossroads.net/app/camps/application_page/camps_payment/camps_payment.controller.js
Ganonside: it's an alias to localhost to trick the payment widget into thinking we're on the crossroads.net domain. Otherwise we'll get a CORS error when running crds-embed locally.

PR #3274, opened 2016-12-09
US6015 - Apply Styles to Horizontal Slider on Stream Page

zarahzachz:
Add style changes so /live/stream more closely matches mockup. Change geolocation buttons to match embed's color scheme better.

+ Added margin-left to "Don't Miss" section (margin-right is still 0 on desktop screens)
+ Change horizontal slider buttons to use glyphs instead of text
+ Style geolocation buttons to use a consistent blue & gray
+ Change width of "Don't Miss" cards to a thinner size (to match mockup)
+ Fix text-cutoff issue on the titles for "Be the Church" section


tcmacdonald:
@zarahzachz @beckysingson This PR needs to be updated to reflect the work in #3275. Please rebase your branch against `feature/US5960-horizontal-scrolling`.

PR #3297, opened 2016-12-14
DE2455 Set width to important on input groups

ilpadre:


ilpadre:
The issue isn't that we can't override the width. The issue is that it is being overridden without our consent. And this would apply to only those elements with the style input-group applied to it, so it wouldn't just be a blanket, "you can't override width anymore" kind of thing. 

I can see the value of creating another class specific to this form, but we use the input-group class to style input elements all over the site. Do they have the same problem? And if they do, do we create a custom class for them as well?

I think we'd still end up using the !important modifier. I think our new class width would get overridden. The real solution is to figure out why it's being overridden in the first place.

PR #3302, opened 2016-12-15
Defect/de2456 room drop down not loading

JoeKer1:


crossroads.net/app/mp_tools/add_event_tool/add_room/equipment_form/equipmentForm.controller.js
```diff
@@ -8,7 +8,7 @@ export default class EquipmentController {
   }
 
   addEquipment() {
-    this.currentEquipment.push({ equipment: { name: null, quantity: 0 } });
+    this.currentEquipment.push({ equipment: { name: { id: 0 }, quantity: 0 } });
```
ilpadre: Name is an object containing an id of 0? I'm confused.

JoeKer1: Yeah you and me both. That is actually what the model looks like when you select equipment in the drop down. The way I implemented equipment it starts with one with id of zero and then I filter those out in the mapping. I didn't think we had time to rework both the front end and backend equipment calls and models

PR #3303, opened 2016-12-15
Defect/de2453 site and phonenumber issues

silbermm:


crossroads.net/app/camps/camps.routes.js
```diff
@@ -75,7 +75,8 @@ export default function CampRoutes($stateProvider) {
         CampsService: 'CampsService',
         $state: '$state',
         getCamperPayment,
-        getCamperFamily
+        getCampInfo,
+        getCamperFamily,
```
Ganonside: Are we doing trailing commas or not? I vaguely remember being told that we weren't going to do those.

PR #3310, opened 2016-12-15
US5967: Add Accounting Company to Congregation filter

eymiha:
Adjust the filter for congregation to include the selected accounting company


IngageKrizJ:
Only change was a Report (.rdl) file, failed build due to Demo DB restore issue, so merging without successful build.

PR #3316, opened 2016-12-16
DE2449 Iframe Resizer Redux

tcmacdonald:
This PR refactors the current implementation of the embedded, inline-giving iframe into a new directive to resolve a race condition related to instantiating the resizer script against DOM elements populated by a content-block.

### DIFF
--- a/crossroads.net/app/live_stream/stream/stream.html
+++ b/crossroads.net/app/live_stream/stream/stream.html
@@ -19,7 +19,7 @@
 
   <div class="digital-program-container">
     <div class="flex__container container">
-      <div class="flex__item digital-program__giving" dynamic-content="$root.MESSAGES.streamingInlineGivingIframe.content | html"></div>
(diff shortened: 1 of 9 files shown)


crossroads.net/app/live_stream/stream/stream.html
```diff
@@ -19,7 +19,7 @@
 
   <div class="digital-program-container">
     <div class="flex__container container">
-      <div class="flex__item digital-program__giving" dynamic-content="$root.MESSAGES.streamingInlineGivingIframe.content | html"></div>
+      <div class="flex__item digital-program__giving" dynamic-content="$root.MESSAGES.streamingInlineGivingEmbed.content | html"></div>
```
tcmacdonald: I created a new content-block solely for testing locally so that I wouldn't screw up anything QA is currently working on in INT. If we're cool w/ the approach in this PR, I'll revert this change (since the streamingInlineGivingIframe already exists in prod) and update the content block to use our new directive instead of it's current implementation.

PR #3324, opened 2016-12-19
Feature/us3065 edit nonapproval fields

Doug-Shannon:
This is complete.  It allows the update of all fields, even ones that force a re-approval.  Next story will finish the process where approvals are invalidated and recreated.


crossroads.net/app/mp_tools/add_event_tool/addEvent.service.js
```diff
@@ -67,10 +68,12 @@
...
             },
-            reminderDays: event.reminderDaysId,
+            reminderDays: {
+              dp_RecordID: event.reminderDaysId
+            },
             sendReminder: event.sendReminder,
-            startTime: new Date(event.startDateTime),
-            endTime: new Date(event.endDateTime),
+            startTime: new Date(event.startDateTime + "-0500"),
+            endTime: new Date(event.endDateTime + "-0500"),
```
Doug-Shannon: I am not so certain this is the correct way to handle the time zone.  The time comes back without it, so I added EST to fix it.  Not sure if we need to update the save to send timezone or what.

PR #3331, opened 2016-12-19
Feature/us3065 edit nonapproval fields

Doug-Shannon:
Updates the Manage Event form to allow edits on the event.


crossroads.net/app/mp_tools/add_event_tool/addEvent.service.js
```diff
@@ -46,6 +46,10 @@
       },
 
       fromEventDto(event) {
+        if (!event.group){
+          event.group = {};
+        }
+        this.origCongregation = event.congregationId;
```
eymiha: I see this getting tucked away, but I'm not seeing what's using it later.

crossroads.net/app/mp_tools/add_event_tool/add_event/addEvent.controller.js
```diff
@@ -71,13 +72,13 @@ export default class AddEventcontroller {
...
 
     return new Date(
-            dateForDate.getFullYear(),
-            dateForDate.getMonth(),
-            dateForDate.getDate(),
-            dateForTime.getHours(),
-            dateForTime.getMinutes(),
-            dateForTime.getSeconds(),
-            dateForTime.getMilliseconds());
+      dateForDate.getFullYear(),
```
eymiha: It seems like this should get tucked away in a utility package...

Doug-Shannon: We refactored the whole manage even tool to ang 1.5 and most of this was carry over.  Had to make some time based decisions on what else would get pulled out.

This may be a good candidate to be stashed somewhere for our next story, I think we may need it.

PR #3342, opened 2016-12-20
US5241 camp payment email

silbermm:
* Created a new route that is called after successful payment (payment-confirmation)
* Created a new endpoint to send the email
* Created email template to send and attached to the Summer Camps program


Gateway/crds-angular/Controllers/API/PaymentController.cs
```diff
@@ -41,69 +41,87 @@ public IHttpActionResult AlreadyPaidDeposit(int invoiceId)
...
+                {
+                    var apiError = new ApiErrorDto("Unable to get payment details", e);
+                    throw new HttpResponseException(apiError.HttpResponseMessage);
+                }
+                
+            });
+            
+        }
+
+        [VersionedRoute(template: "payment/{paymentId}/confirmation", minimumVersion: "1.0.0")]
```
Doug-Shannon: I might be wrong, but don't you need the [Requires Authorization] attribute for these to work for testing in swagger?

crossroads.net/app/camps/camps.service.js
```diff
@@ -157,6 +158,14 @@ class CampsService {
       });
   }
 
+  sendPaymentConfirmation(invoiceId, paymentId, eventId, contactId) {
+    return this.paymentConfirmationResource.save({ contactId, eventId, invoiceId, paymentId }, {}).$promise
+      .then(() => {
+        this.initializeCampData();
+        this.initializeCamperData();
+      });
```
Doug-Shannon: Is there an error to handle on save here?

Gateway/crds-angular/Controllers/API/PaymentController.cs
silbermm: I'm not sure... I've not heard that, but I also don't typically use Swagger. I'll add just in case.

crossroads.net/app/camps/camps.service.js
silbermm: I think we are handling it [here](https://github.com/crdschurch/crds-angular/pull/3342/files/3587b0f41163ed359e241c17beffeaec3c6e96d0#diff-0269a437d5bb823c3ef423b40c68d1e8R67) where we consume this service by calling `.finally()` . Essentially we always want to route back to the dashboard regardless of success or failure.

PR #3344, opened 2016-12-21
Defect/de2486 stream layout on iphone 6 broken

zarahzachz:
Updating layout - page layout now relies on Bootstrap grid and smaller components are driven by flexbox.

> But on browserstack, iphone 6/plus /stream layout is messed up. The inline giving widget and the don't miss section does not look correct


zarahzachz:
Amendment to PR: Fix to articles in horizontal slider - they have to be able to expand when content is added, but also all have the same height.

PR #3361, opened 2016-12-27
Feature/DE2550 add redirect to apache config template file

aiqingyuyan:


aiqingyuyan:
Did a build in runway-1 environment, all checks passed

k8tied1:
1. Please rename this PR with the correct info. I'm assuming it should be DE2550.  2. Where is the CRDS_REDIRECT_SERVER_NAME set? I looked through the team city config and couldn't find it. 3. I'm still working on a comparison between the template and prod.

k8tied1:
Differences between prod and template:
in 80 and 443 sections
In the redirect, there's a trailing / in prod but not in the template 
There are no RewriteCond lines at all in prod
Missing Include /var/www/crossroads.net/apache_site.conf
apache logs don't have server name in production - they are just error.log and access.log

Current in prod cache headers (/assets) is in an if statement
  # Cache Headers
  <ifmodule mod_headers.c>
 
443  
current in prod there's no ProxyPass for 443

PR #3364, opened 2016-12-28
US6100 add flag to Participants for host approval

sseissiger:


CI/SQL/20161228_112000_US6100-appoved-host-flag.sql
```diff
@@ -0,0 +1,13 @@
+USE [MinistryPlatform]
+GO
+
+SET ANSI_NULLS ON
+GO
+
+SET QUOTED_IDENTIFIER ON
+GO
+
+ALTER TABLE [dbo].[Participants]
```
andycanterbury: We typically test to make sure the table exists before modifying it. Although, if Participants is missing, we are in BIG trouble. Also, check that the column doesn't already exist before trying to add. It makes the scripts safe to rerun and less likely to fail and break a build.

PR #3365, opened 2016-12-28
Feature/us4980 disable apache directory listing

aiqingyuyan:


k8tied1:
I thought that for the apache directory listing we were hoping to find a more global answer. I may be wrong (and please correct me if I am) but it looks like this will be done seperately for each site. 

I did some research here https://httpd.apache.org/docs/current/howto/htaccess.html and found 'In general, you should only use .htaccess files when you don't have access to the main server configuration file.'

PR #3375, opened 2017-01-03
Adding an ID to the More tab on the profile page

candrews-thrive:
to make finding the control easier in selenium

### DIFF
--- a/crossroads.net/app/profile/profile.html
+++ b/crossroads.net/app/profile/profile.html
@@ -26,7 +26,7 @@
       <ui-view ng-if="tab.active"></ui-view>
     </tab>
     <li role="tab" class="dropdown hidden-md hidden-lg hidden-sm dropdown-toggle" dropdown is-open="status.isopen">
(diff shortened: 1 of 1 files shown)


crossroads.net/app/profile/profile.html
```diff
@@ -26,7 +26,7 @@ <h3 class="profile-name">{{profile.profileData.person.nickName}} {{profile.profi
       <ui-view ng-if="tab.active"></ui-view>
     </tab>
     <li role="tab" class="dropdown hidden-md hidden-lg hidden-sm dropdown-toggle" dropdown is-open="status.isopen">
-      <a dropdown-toggle ng-disabled="disabled">More <span class="caret"></span></a>
+      <a id="MoreTab" dropdown-toggle ng-disabled="disabled">More <span class="caret"></span></a>
```
hookom: Can you make this camel case ("moreTab")? I think that's the convention used with other ids in the app.

candrews-thrive: Done

PR #3377, opened 2017-01-04
Feature/us5946 reapprove key edits

Doug-Shannon:


Gateway/MinistryPlatform.Translation/Repositories/MinistryPlatformRestRepository.cs
```diff
@@ -214,6 +215,58 @@ private static string FormatStoredProcParameters(Dictionary<string, object> para
...
+
+
+            var url = $"/tasks/{taskID}?";
+            var request = new RestRequest(baseURL, Method.GET);
+            AddAuthorization(request);
+
+            var response = _ministryPlatformRestClient.Execute(request);
+            _authToken.Value = null;
+            response.CheckForErrors($"Error searching for task");
+        }
```
ilpadre: Comments say that this added code is not tested and is not currently used. When will this happen? Does it need to be in here now?

Doug-Shannon: yeah, it is relevant to a possible next story.  We can chat about it.

PR #3388, opened 2017-01-04
Feature release/feature/us6013 cancel event

JoeKer1:


crossroads.net/jsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+		"allowSyntheticDefaultImports": true
+	},
+	"exclude": [
+		"node_modules",
+		"bower_components",
+		"jspm_packages",
+		"tmp",
+		"temp"
+	]
+}
```
ilpadre: Why were these changes made?

JoeKer1: Ohh, those shouldn't have gotten in there, that's config for vscode.  well maybe they should be in there.

PR #3399, opened 2017-01-05
DE2583: Rick is old. That is all.

samueljmello:


crossroads.net/core/login/login_controller.js
```diff
@@ -105,10 +104,11 @@
 
            500);
           } else if ($scope.loginCallback) {
+            $scope.processing = false;
```
vchtelmakh: Don't need to change, but I think it's a good convention to use `isProcessing`  /  `isWhatever` for bools, especially when we aren't using typescript.

PR #3404, opened 2017-01-06
DE2581 Modified Stored procedure and CampService

Ganonside:
Now checking group participant end date and group end date when
determining eligibility for camp.

Also doesn't check eligibility if the application has already been
started.


Gateway/crds-angular.test/Services/CampServiceTest.cs
```diff
@@ -470,7 +471,6 @@ public void shouldGetCampFamilyNotHeadSignedUp()
             _contactService.Setup(m => m.GetMyProfile(token)).Returns(myContact);
             _contactService.Setup(m => m.GetHouseholdFamilyMembers(myContact.Household_ID)).Returns(getFakeHouseholdMembers(myContact, false, "Adult Child"));
             _apiUserRepository.Setup(m => m.GetToken()).Returns(apiToken);
-            _groupRepository.Setup(m => m.IsMemberOfEventGroup(myContact.Contact_ID, eventId, apiToken)).Returns(true);
```
andycanterbury: How come you can remove this mock when the code still calls IsMemberOfEventGroup? Is there a path we aren't testing?

Ganonside: that specific test doesn't make that call. But this call is tested in other unit tests

PR #3406, opened 2017-01-06
US5814 use auth this time

andycanterbury:
Also, count the declines because they zero out.


Gateway/crds-angular/Services/PaymentService.cs
```diff
@@ -270,7 +270,7 @@ public DonationBatchDTO GetPaymentBatch(int batchId)
         public void UpdateInvoiceStatusAfterDecline(int invoiceId)
         {
             var payments = _paymentRepository.GetPaymentsForInvoice(invoiceId);
-            payments = payments.Where(p => p.PaymentStatus != _declinedPaymentStatus).ToList();
+            //payments = payments.Where(p => p.PaymentStatus != _declinedPaymentStatus).ToList();
```
silbermm: can we just remove this line, or do you want to keep it here?

PR #3417, opened 2017-01-06
Feature release/defect/de2597-Api user create edit

Doug-Shannon:


JoeKer1:
Do see anything with event groups, hopefully that one just works. but based on the permissions i dont' think it will

JoeKer1:
Also labels and milestones

PR #3434, opened 2017-01-09
US5306 Impersonation

richardbarker:


samueljmello:
DO NOT MERGE UNTIL AFTER 12:30 PM on 1/9/2017

crossroads.net/spec-es6/core/impersonate/impersonate.controller.spec.js
```diff
@@ -0,0 +1,78 @@
...
+    httpBackend.flush();
+
+    expect(fixture.error).toBe(false);
+    expect(rootScope.userid).toBe(mockResponse.userId);
+    expect(rootScope.username).toBe(mockResponse.username);
+    expect(rootScope.email).toBe(mockResponse.userEmail);
+
+  });
+
+  it('should set fail and keep user credentials on unsuccessful API call', () => {
```
mjvcallibrity: Looks like you meant `should fail and keep...` instead of `should set fail and keep...`

richardbarker: It is setting a flag indicating failure - so I think this is appropriately named

mjvcallibrity: Got it. Thanks!

Gateway/crds-angular/Controllers/API/SubscriptionsController .cs
```diff
@@ -14,7 +14,7 @@ public class SubscriptionsController : MPAuth
         private readonly MPInterfaces.IAuthenticationRepository _authenticationService;
         private readonly ISubscriptionsService _subscriptionService;
 
-        public SubscriptionsController(ISubscriptionsService subscriptionService, MPInterfaces.IAuthenticationRepository authenticationService)
+        public SubscriptionsController(ISubscriptionsService subscriptionService, MPInterfaces.IAuthenticationRepository authenticationService, IUserImpersonationService userImpersonationService) : base(userImpersonationService)
```
richardbarker: It was like that before, but good catch, I removed the space

Gateway/crds-angular/Security/MPAuth.cs
```diff
@@ -49,23 +57,38 @@ protected IHttpActionResult Authorized(Func<string, IHttpActionResult> actionWhe
...
                 if (Request.Headers.TryGetValues("RefreshToken", out refreshTokens) && refreshTokens.Any())
                 {
                     var authData = AuthenticationRepository.RefreshToken(refreshTokens.FirstOrDefault());
                     if (authData != null)
                     {
                         authorized = authData["token"].ToString();
                         var refreshToken = authData["refreshToken"].ToString();
-                        var result = new HttpAuthResult(actionWhenAuthorized(authorized), authorized, refreshToken);
+                        IHttpActionResult result = null;
+                        if (impersonate)
```
IngageKrizJ: Style thing - make sure to use braces around these code blocks (even though only one line).

Gateway/crds-angular/Security/MPAuth.cs
```diff
@@ -49,23 +57,38 @@ protected IHttpActionResult Authorized(Func<string, IHttpActionResult> actionWhe
...
+                            result = new HttpAuthResult(actionWhenAuthorized(authorized), authorized, refreshToken);
                         return result;
                     }
                 }
 
                 authorized = Request.Headers.GetValues("Authorization").FirstOrDefault();   
                 if (authorized != null && (authorized != "null" || authorized != ""))
                 {
-                    return actionWhenAuthorized(authorized);
+                    if (impersonate)
```
IngageKrizJ: Same here - enclose code blocks in curly braces.

crossroads.net/core/impersonate/impersonate.controller.js
```diff
@@ -0,0 +1,75 @@
...
+        this.setCurrentUser(response);
+      })
+      .error(() => {
+        this.processing = false;
+        this.error = true;
+        this.storeImpersonateDetails(false);
+      });
+    };
+
+    this.stopImpersonating = () => {
```
IngageKrizJ: Should something like this get invoked when a user explicitly logs out, or when a session times out?  Since impersonation stuff is stored on the $rootScope, it will still persist after a logout (i think), and logging back in with a different user that is not allowed to impersonate might give unexpected results.

Gateway/crds-angular/Security/MPAuth.cs
samueljmello: There is. It's just on the next line. It's C# with that automatic line returning for brackets.

crossroads.net/core/impersonate/impersonate.controller.js
samueljmello: @IngageKrizJ All that method does is change the stored user properties already in use. During a refresh (or logout) those properties are cleared. All properties used for validation are updated/cleared on log out / log in with new values. Furthermore, server-side validation is completed as well to avoid an advanced client attack that would manually adjust properties and then attempt to impersonate. Does that address your concerns?

Gateway/crds-angular/Security/MPAuth.cs
samueljmello: There is. It's just on the next line. It's C# with that automatic line returning for brackets.

samueljmello: @IngageKrizJ see above comment.

Gateway/crds-angular/Security/MPAuth.cs
IngageKrizJ: I don't see curly braces...
image

crossroads.net/core/impersonate/impersonate.controller.js
IngageKrizJ: Yea, it wasn't a huge concern, just wanting to make sure we cleaned up things if needed.  Sounds like it is probably ok.

PR #3436, opened 2017-01-09
Defect/de2571 single digit bday

Ganonside:
Keeping it at 2 digits in the month and day sections. Placeholder now showing


crossroads.net/app/camps/application_page/camper_info/camper_info.controller.js
```diff
@@ -21,6 +21,8 @@ class CamperInfoController {
     this.submitting = true;
 
     if (this.infoForm.$valid) {
+      console.log(this.infoForm);
```
silbermm: remove the `console.log`

PR #3437, opened 2017-01-09
The report is NEW, not existing, so it needs entries on the reports t…

IngageRitterS:
…able

### DIFF
--- a/CI/SQL/20170109-1313600_U6297-AddReportEntires.sql
+++ b/CI/SQL/20170109-1313600_U6297-AddReportEntires.sql
@@ -1,0 +1,43 @@
+USE [MinistryPlatform]
+GO
+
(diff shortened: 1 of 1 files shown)


CI/SQL/20170109-1313600_U6297-AddReportEntires.sql
```diff
@@ -0,0 +1,43 @@
...
+           ,[Report_Path]
+           ,[Pass_Selected_Records]
+           ,[Pass_LinkTo_Records]
+           ,[On_Reports_Tab])
+     VALUES
+           (308
+		   ,'Contribution Statements'
+           ,'Contribution Statements'
+           ,'/MPReports/crossroads/CRDS Statement Columns'
+           ,1
```
charlie-retzler: Busted... This should be 0

PR #3462, opened 2017-01-12
DE2574 extra commas

silbermm:
* frontend was passing null values to backend
* filtered out all null values before saving to MP


danrye:
I asked Matt to merge, I was waiting on the build to finish and had a meeting.

PR #3470, opened 2017-01-12
replace ApiVersioning code with NuGet package

eymiha:
Removes the existing code from the project and includes the Crossroads.ApiVersioning package from NuGet


Gateway/crds-angular/packages.config
```diff
@@ -31,6 +32,7 @@
   <package id="Microsoft.Owin.Security.Twitter" version="3.0.1" targetFramework="net45" />
   <package id="Microsoft.Web.Infrastructure" version="1.0.0.0" targetFramework="net45" />
   <package id="Newtonsoft.Json" version="9.0.1" targetFramework="net45" />
+  <package id="NUnit3TestAdapter" version="3.6.0" targetFramework="net45" />
```
IngageKrizJ: I see this is a dependency declared in the nuspec for Crossroads.ApiVersioning, but it is really only needed at development time, we don't really want projects to have to include it at runtime.  I think if we take it out of the declared dependencies in the nuspec, this will go away.

eymiha: NUnit dependency removed.

log4net upgraded to 2.0.7 in crds-angular projects to eliminate mismatch.

Passed local smoke test.

PR #3513, opened 2017-01-18
Feature release/onsite groups

JoeKer1:


jpc11623:
This was previously CR'ed interally on Scrummy Bears. I was asked to merge it as part of their runway process.

PR #3584, opened 2017-02-10
DE2734 Resolve inconsistent sign-in behavior

richiebrees:
Code change routes user to site root instead of to the profile view.  Inconsistent behavior was due to session_service logging out the user during the auth check occurring every 3 seconds.

### DIFF
--- a/crossroads.net/core/login/login_controller.js
+++ b/crossroads.net/core/login/login_controller.js
@@ -32,12 +32,7 @@
     var vm = this;
     vm.path = ImageService.ProfileImageBaseURL + vm.contactId;
     vm.defaultImage = ImageService.DefaultProfileImage;
-
-    // called during unit tests to ensure tested properties and functions are available
-    vm.setupTesting = function(){
-      vm.navigateToHome = navigateToHome;
-      vm.login = $scope.login;
-    };
+    vm.navigateToHome = navigateToHome;
 
     $scope.loginShow = false;
     $scope.newuser = User;
@@ -74,14 +69,13 @@
       return;
     };
 
-    function navigateToHome()
-    {
+    function navigateToHome() {
       $state.go('content', { link: '/' });
-    }    
+    }
 
     $scope.login = function() {
       if (($scope.credentials === undefined) ||
-          ($scope.credentials.username === undefined ||
+        ($scope.credentials.username === undefined ||
           $scope.credentials.password === undefined)) {
         $scope.pending = true;
         $scope.loginFailed = false;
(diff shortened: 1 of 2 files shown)


crossroads.net/core/login/login_controller.js
```diff
@@ -77,6 +89,7 @@
       } else {
         $scope.processing = true;
         AuthService.login($scope.credentials).then(function(user) {
+
```
mjvcallibrity: Go ahead and remove this whitespace.

crossroads.net/core/login/login_controller.js
```diff
@@ -67,6 +74,11 @@
       return;
     };
 
+    function navigateToHome()
+    {
```
mjvcallibrity: General practice is to include the `{` after the `()`. Don't know if it is considered symantic JS, but that is the practice Crossroads has been following.

crossroads.net/spec-core/login/login.controller.spec.js
```diff
@@ -0,0 +1,88 @@
...
+
+    LoginController.setupTesting();
+
+  }));
+
+  it('controller should be defined', function () {
+    expect(LoginController).toBeDefined();
+  });
+
+  it('function navigateToHome routes to homepage', function () {
```
mjvcallibrity: Crossroads has a practice of using `should` in a unit test descriptions, so instead of reading "function navigateToHome routes to homepage", it would read "function navigateToHome should route to homepage".

I know. Really nitpicky, but in general Architecture appreciates consistency throughout the system.

crossroads.net/spec-core/login/login.controller.spec.js
```diff
@@ -0,0 +1,88 @@
...
+    spyOn($state, 'go');
+    LoginController.navigateToHome();
+    expect($state.go).toHaveBeenCalledWith('content', { link: '/' });
+  });
+
+  it('did route to homepage after signing in and no redirect parameters were passed', function () {
```
mjvcallibrity: Can substitute "did" with "should".

crossroads.net/core/login/login_controller.js
```diff
@@ -32,6 +32,13 @@
     var vm = this;
     vm.path = ImageService.ProfileImageBaseURL + vm.contactId;
     vm.defaultImage = ImageService.DefaultProfileImage;
+
+    // called during unit tests to ensure tested properties and functions are available
```
mjvcallibrity: Feels like this kind of helper method should be contained in the test file itself. Is that possible? Would you be able to call this code in the test file like such:

`LoginController.vm.navigateToHome = $state.go('content', { link: '/' });
LoginController.vm.login = $scope.login`

PR #3588, opened 2017-02-13
Updated communication_ID so the correct email would be sent

lippsm:


hookom:
Can you put that ID into a variable and add a comment above the variable describing why we have to specify it?  Seems like something we'll forget about and run into copy/paste errors later.

PR #3595, opened 2017-02-14
DE2723 Handle other households properly

silbermm:
Changed translation layer to look at other members from the Contacts_Households table instead of Other_Households.

### DIFF
--- a/Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
+++ b/Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
@@ -206,20 +206,41 @@ namespace MinistryPlatform.Translation.Repositories
             }).ToList();
             return family;
         }
-
-        public List<MpHouseholdMember> GetOtherHouseholdMembers(int contactId)
+        
+        /// <summary>
+        /// Get the Other/Former members of a particular household
+        /// </summary>
+        /// <param name="householdId">the householdId of the household where you want the other members</param>
+        /// <returns> a list of MpHouseholdMember </returns>
+        public List<MpHouseholdMember> GetOtherHouseholdMembers(int householdId)
         {
             var token = ApiLogin();
-            var householdMembers = new List<MpHouseholdMember>();
-            var otherHouseholds = _ministryPlatformService.GetSubpageViewRecords("OtherHouseholds", contactId, token);
-            foreach (var house in otherHouseholds)
+            var filter = $"Contact_Households.Household_ID = {householdId}";
+            var columns = new List<string>
             {
-                var houseId = (int) house["Household_ID"];
-                householdMembers.AddRange(GetHouseholdFamilyMembers(houseId));
-            }
+                "Contact_Households.Contact_ID",
+                "Contact_Households.Household_ID",
+                "Contact_Households.Household_Position_ID",
+                "Household_Position_ID_Table.Household_Position",
+                "Contact_ID_Table.First_Name",
+                "Contact_ID_Table.Nickname",
+                "Contact_ID_Table.Last_Name",
+                "Contact_ID_Table.Date_of_Birth",
+                "Contact_ID_Table.__Age"
+            };
+            var result = _ministryPlatformRest.UsingAuthenticationToken(token).Search<MpContactHousehold>(filter, columns);
+            var householdMembers = result.Select((hm) => new MpHouseholdMember
+            {
+                Age = hm.Age ?? 0,
+                ContactId = hm.ContactId,
+                DateOfBirth = hm.DateOfBirth ?? new DateTime(),
+                FirstName = hm.FirstName,
+                LastName = hm.LastName,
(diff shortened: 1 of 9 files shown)


Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -206,20 +206,41 @@ public List<MpHouseholdMember> GetHouseholdFamilyMembers(int householdId)
...
         {
             var token = ApiLogin();
-            var householdMembers = new List<MpHouseholdMember>();
-            var otherHouseholds = _ministryPlatformService.GetSubpageViewRecords("OtherHouseholds", contactId, token);
-            foreach (var house in otherHouseholds)
-            {
-                var houseId = (int) house["Household_ID"];
-                householdMembers.AddRange(GetHouseholdFamilyMembers(houseId));
-            }
+            var filter = $"Contact_Households.Household_ID = {householdId}";
```
andycanterbury: Contact_Households has an End_Date field. I think we should account for that in the filter too.

Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -206,20 +206,41 @@ public List<MpHouseholdMember> GetHouseholdFamilyMembers(int householdId)
...
+                "Contact_ID_Table.First_Name",
+                "Contact_ID_Table.Nickname",
+                "Contact_ID_Table.Last_Name",
+                "Contact_ID_Table.Date_of_Birth",
+                "Contact_ID_Table.__Age"
+            };
+            var result = _ministryPlatformRest.UsingAuthenticationToken(token).Search<MpContactHousehold>(filter, columns);
+            var householdMembers = result.Select((hm) => new MpHouseholdMember
+            {
+                Age = hm.Age ?? 0,
```
andycanterbury: Does putting 0 in for age cause any issues elsewhere? I know sometimes we assume if we don't have an age for someone then they are an adult. Does a 0 throw off that code?

Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -206,20 +206,41 @@ public List<MpHouseholdMember> GetHouseholdFamilyMembers(int householdId)
...
+                ContactId = hm.ContactId,
+                DateOfBirth = hm.DateOfBirth ?? new DateTime(),
```
andycanterbury: Same here. Does having a new datetime as a birthdate mess up any code elsewhere in the system? Would a nullable be better here?

Gateway/crds-angular/Services/CampService.cs
```diff
@@ -198,7 +196,7 @@ public void SaveCamperEmergencyContactInfo(List<CampEmergencyContactDTO> emergen
         {
             var loggedInContact = _contactRepository.GetMyProfile(token);
             var family = _contactRepository.GetHouseholdFamilyMembers(loggedInContact.Household_ID);
-            family.AddRange(_contactRepository.GetOtherHouseholdMembers(loggedInContact.Contact_ID));
+            family.AddRange(_contactRepository.GetOtherHouseholdMembers(loggedInContact.Household_ID));
```
andycanterbury: Not your fault and I don't expect it to be cleaned up now, but this needs to be DRYed up. We do this getting family and other family over and over and over again in this service...

Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -17,7 +17,7 @@ namespace MinistryPlatform.Translation.Repositories
     public class ContactRepository : BaseRepository, IContactRepository
     {
         private readonly int _addressesPageId;
-        private readonly IConfigurationWrapper _configurationWrapper;
+        //private readonly IConfigurationWrapper _configurationWrapper;
```
andycanterbury: Remove if no longer needed.

Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -35,7 +35,7 @@ public ContactRepository(IMinistryPlatformService ministryPlatformService, IAuth
             : base(authenticationService, configuration)
         {
             _ministryPlatformService = ministryPlatformService;
-            _configurationWrapper = configuration;
+            //_configurationWrapper = configuration;
```
andycanterbury: Remove if not needed.

Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -35,7 +35,7 @@ public ContactRepository(IMinistryPlatformService ministryPlatformService, IAuth
             : base(authenticationService, configuration)
```
andycanterbury: If the configuration wrapper isn't needed anymore, do we need to remove something here?

PR #3603, opened 2017-02-15
Feature/us6488 display detail for self

JoeKer1:


Gateway/MinistryPlatform.Translation/Models/MpEvent.cs.orig
```diff
@@ -0,0 +1,100 @@
+﻿using System;
```
plachmann: looks like a copy of a file made it in. please remove.

crossroads.net/app/mp_tools/add_event_tool/add_room/equipment_form/equipment_form.html.orig
```diff
@@ -0,0 +1,109 @@
+<div ng-show="equipment.equipmentLookup.length < 1"> There is no equipment to reserve at this site </div>
```
plachmann: another .orig file to remove

PR #3606, opened 2017-02-16
US6283-DetermineInitialLocation

plachmann:


Gateway/MinistryPlatform.Translation/Repositories/FinderRepository.cs
```diff
@@ -43,6 +51,21 @@ public FinderPinDto GetPinDetails(int participantId)
             return pinDetails;
         }
 
+        public string GetIpForRemoteUser()
+        {
+            var ip = "";
 
+            {
```
eymiha: The block from 58 to 67 looks like it's not needed... but it's not hurting anything.

Gateway/crds-angular/Controllers/API/FinderController.cs
```diff
@@ -38,5 +39,42 @@ public IHttpActionResult GetPinDetails([FromUri]int participantId)
...
+        {
+            try
+            {
+                var participantId = _finderService.GetParticipantIdFromContact(contactId);
+                var list = _finderService.GetPinDetails(participantId);
+                return Ok(list);
+            }
+            catch (Exception ex)
+            {
+                var apiError = new ApiErrorDto("Get Leaders Groups Failed", ex);
```
eymiha: Should the message be "Get Pin Details by Contact Failed"?

PR #3607, opened 2017-02-16
Feature/us6820 create host statuses

eymiha:
- Creates and populates a Host_Statuses lookup table,
- Drops the Approved_Host column from the Participant's table, and
- Adds the Hist_Status_ID column to the Participant's table.


eymiha:
The Approved_Host column will not be deleted from Participants until we accommodate it's current users somehow.

PR #3614, opened 2017-02-17
DE2785-make sure lat and long are returned

plachmann:


Gateway/crds-angular/Services/AddressService.cs
```diff
@@ -50,6 +50,8 @@ public void SetGeoCoordinates(AddressDTO address)
                 var coords = _addressGeocodingService.GetGeoCoordinates(address);
                 address.Longitude = coords.Longitude;
                 address.Latitude = coords.Latitude;
+                var mpAddress = AutoMapper.Mapper.Map<MpAddress>(address);
```
sseissiger: Assuming you tested, but I'm not sure these 2 lines are needed.  I will merge, but I'll look at closer/test out variations locally Monday.

PR #3616, opened 2017-02-17
Removing dependent data from KC & Superbowl test groups

candrews-thrive:
* For '(t) KidsClub Oakley Group' the script is now removing opportunities
attach to it.
* For '(t) Superbowl Oakley Group' add a line to remove any event participant attach to it.


lippsm:
Looks good.

PR #3625, opened 2017-02-18
Feature/us6612 camp registration report

Ganonside:


CI/SQL/20170216_095900_US6612-Camp-registration-report.sql
```diff
@@ -0,0 +1,176 @@
+-- ================================================
```
silbermm: delete this metadata

CI/SQL/20170216_095900_US6612-Camp-registration-report.sql
```diff
@@ -0,0 +1,176 @@
...
+SET ANSI_NULLS ON
+GO
+SET QUOTED_IDENTIFIER ON
+GO
+-- =============================================
+-- Author:		Jon Horner
+-- Create date: 2/16/2017
+-- Description:	Gets data for camp registration report
+-- =============================================
+CREATE PROCEDURE report_CRDS_Camp_Registration
```
silbermm: we should be checking for the existence of this procedure and updating if it exists or creating new if doesn't exist, this way it's re-runnable and won't fail if the procedure exists

PR #3630, opened 2017-02-20
DE2803-IPAddressLocation

plachmann:


Gateway/crds-angular/Controllers/API/FinderController.cs
```diff
@@ -74,13 +74,13 @@ public IHttpActionResult GetPinDetailsByContact([FromUri]int contactId)
 
         [ResponseType(typeof(AddressDTO))]
         [VersionedRoute(template: "finder/pinbyip", minimumVersion: "1.0.0")]
```
vchtelmakh: Needs to be:

    ```[VersionedRoute(template: "finder/pinbyip/{ipAddress}", minimumVersion: "1.0.0")]```

This is why we got CORS.

PR #3632, opened 2017-02-20
DE2732 Serial Camper Syndrome

silbermm:
* update FormSubmisssionRepository to do the correct action based on the
  data passed in.
* write tests to validate behavior
* update all places where FormSubmissionRepository method is called to
  include EventID in the Object

### DIFF
--- a/Gateway/MinistryPlatform.Translation.Test/Services/FormSubmissionRepositoryTest.cs
+++ b/Gateway/MinistryPlatform.Translation.Test/Services/FormSubmissionRepositoryTest.cs
@@ -1,0 +1,292 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using Crossroads.Web.Common.Configuration;
+using Crossroads.Web.Common.MinistryPlatform;
+using Crossroads.Web.Common.Security;
+using MinistryPlatform.Translation.Models;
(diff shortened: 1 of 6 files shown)


Gateway/MinistryPlatform.Translation.Test/Services/FormSubmissionRepositoryTest.cs
```diff
@@ -0,0 +1,292 @@
...
+                {"Contact_ID", formResponse.ContactId},
+                {"Opportunity_ID",  formResponse.OpportunityId },
+                {"Opportunity_Response", formResponse.OpportunityResponseId},
+                {"Pledge_Campaign_ID", formResponse.PledgeCampaignId},
+                {"Event_ID", formResponse.EventId}
+            };
+
+            var searchString = $"Contact_ID={formResponse.ContactId} AND Form_ID={formResponse.FormId} AND Event_ID={formResponse.EventId}";
+            const string selectColumns = "Form_Response_ID";
+            _ministryPlatformRestRepository.Setup(m => m.UsingAuthenticationToken(apiToken)).Returns(_ministryPlatformRestRepository.Object);
```
andycanterbury: You could probably put this in the setup unless there is a test that doesn't use authentication.

PR #3635, opened 2017-02-20
session_service redirectIfNeeded logic in new event loop

richiebrees:
Corrects timing/eventing issue in session_service where "redirectIfNeeded" not working correctly from mobile device or desktop going directly to "/signout."

### DIFF
--- a/crossroads.net/core/services/session_service.js
+++ b/crossroads.net/core/services/session_service.js
@@ -140,7 +140,7 @@
     // TODO: Get this working to DRY up login_controller and register_controller
     vm.redirectIfNeeded = ($injectedState) => {
       $timeout(() => {
-        // timeout ensures processing occurs on next 'loop'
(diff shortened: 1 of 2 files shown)


crossroads.net/core/services/session_service.js
```diff
@@ -139,16 +139,19 @@
...
     vm.redirectIfNeeded = ($injectedState) => {
-      if (vm.hasRedirectionInfo()) {
-        const url = vm.exists('redirectUrl');
-        const params = vm.exists('params');
-        vm.removeRedirectRoute();
-        if (params === undefined) {
-          $injectedState.go(url);
-        } else {
-          $injectedState.go(url, JSON.parse(params));
+      $timeout(function() {
```
richardbarker: Could be use an arrow function to be consistent:
$timeout(() => {

PR #3637, opened 2017-02-21
DE2809 - don't go to Lat/Long of 0,0

plachmann:


Gateway/crds-angular/Controllers/API/FinderController.cs
```diff
@@ -59,7 +59,7 @@ public IHttpActionResult GetPinDetailsByContact([FromUri]int contactId)
             {
                 var participantId = _finderService.GetParticipantIdFromContact(contactId);
                 var pin = _finderService.GetPinDetails(participantId);
-                if (pin.Address.Latitude == null || pin.Address.Longitude == null)
+                if (pin.Address.Latitude == null || pin.Address.Longitude == null || (pin.Address.Latitude==0 && pin.Address.Longitude==0))
```
vchtelmakh: We should probably use strict equality here.

PR #3648, opened 2017-02-23
DE2732 Serial Camper Syndrome

silbermm:
* missed some key methods that were getting individual form response
  answers.


Gateway/MinistryPlatform.Translation.Test/Services/FormSubmissionRepositoryTest.cs
```diff
@@ -287,6 +287,60 @@ public void ShouldFindAFormResponseByContactIdOnly()
...
+
+            var searchString = $"Form_ID={formId} AND Contact_ID={contactId} AND Event_ID={eventId}";
+            var answersSearch = $"Form_Response_ID={responseId}";
+
+            _ministryPlatformRestRepository.Setup(m => m.UsingAuthenticationToken(apiToken)).Returns(_ministryPlatformRestRepository.Object);
+            _ministryPlatformRestRepository.Setup(m => m.Search<MpFormResponse>(searchString, null as string, null as string, false)).Returns(new List<MpFormResponse> {new MpFormResponse {FormResponseId = responseId } });
+            _ministryPlatformRestRepository.Setup(m => m.Search<MpFormAnswer>(answersSearch, null as string, null as string, false)).Returns(new List<MpFormAnswer>());
+
+            _fixture.GetFormResponse(formId, contactId, eventId);
+            _ministryPlatformRestRepository.VerifyAll();
```
andycanterbury: Is there any value in adding an assert to test that you got the response you expected? Since we are mocking the call I can understand if there isn't.

Gateway/MinistryPlatform.Translation.Test/Services/FormSubmissionRepositoryTest.cs
```diff
@@ -287,6 +287,60 @@ public void ShouldFindAFormResponseByContactIdOnly()
...
+
+            var searchString = $"Form_ID={formId} AND Contact_ID={contactId}";
+            var answersSearch = $"Form_Response_ID={responseId}";
+
+            _ministryPlatformRestRepository.Setup(m => m.UsingAuthenticationToken(apiToken)).Returns(_ministryPlatformRestRepository.Object);
+            _ministryPlatformRestRepository.Setup(m => m.Search<MpFormResponse>(searchString, null as string, null as string, false)).Returns(new List<MpFormResponse> { new MpFormResponse { FormResponseId = responseId } });
+            _ministryPlatformRestRepository.Setup(m => m.Search<MpFormAnswer>(answersSearch, null as string, null as string, false)).Returns(new List<MpFormAnswer>());
+
+            _fixture.GetFormResponse(formId, contactId, null);
+            _ministryPlatformRestRepository.VerifyAll();
```
andycanterbury: Same here. Should we at least assert that we get the right type of thing back?

Gateway/MinistryPlatform.Translation.Test/Services/MinistryPlatformRestServiceIntegrationTest.cs
```diff
@@ -539,6 +539,23 @@ public void findOtherHouseholdMembers()
...
+            {
+                "Form_Response_Answer_ID",
+                "Form_Field_ID",
+                "Form_Response_Answers.Form_Response_ID",
+                "Response",
+                "Form_Response_Answers.Opportunity_Response",
+                "Event_Participant_ID"
+            };
+
+            _fixture.UsingAuthenticationToken(_authToken).Search<MpFormAnswer>(searchString, columns);
```
andycanterbury: I know we don't run integration tests, but this one doesn't assert anything. I'm guessing it was used for your own testing. Can we make it actually test something or get rid of it?

Gateway/MinistryPlatform.Translation/Repositories/FormSubmissionRepository.cs
```diff
@@ -171,12 +171,12 @@ public int SubmitFormResponse(MpFormResponse form)
             return responseId;
         }
 
-        public string GetFormResponseAnswer(int formId, int contactId, int formFieldId)
+        public string GetFormResponseAnswer(int formId, int contactId, int formFieldId, int? eventId)
```
andycanterbury: The interface says eventId should be defaulted to null. That would aso match the GetFormResponse method below.

Gateway/MinistryPlatform.Translation.Test/Services/MinistryPlatformRestServiceIntegrationTest.cs
silbermm: If it doesn't throw an exception, it passes, but I can remove it if it bothers you

PR #3649, opened 2017-02-23
Defect/de2816 camp registration report columns

andycanterbury:


silbermm:
**Several formatting issues with the report:**
* The Birthdate is currently being shown as YYYY-MM-DD but should be shown as MM/DD/YYYY
* The column labels for Additional Emergency Contact should have `Additional` spelled out instead of `Add'l`

Also the event start/end dates are missing underneath the title of the report

PR #3675, opened 2017-03-02
Defect/de2869 alphabetical cities

silbermm:
This fixes both DE2869 and DE2870


crossroads.net/app/go_volunteer/organizations/organizations.controller.js
```diff
@@ -56,14 +56,23 @@ export default class OrganizationsController {
     return this.cities.map((city) => {
       const name = { name: `${city.city}, ${city.state}` };
       return Object.assign(name, city);
```
Ganonside: should be `Object.assign(city, name);`

PR #3677, opened 2017-03-02
US7077 - GO Local anywhere profile confirmation

tallented:
Associated CMS pull request:  https://github.com/crdschurch/SS-CMS/pull/257

URL:  /go-local/3/crossroads/Cleveland/564/confirm

* Creates component and route
* Use content block for informative text
* Display project card and project leader's email address

image

### DIFF
--- a/crossroads.net/app/go_volunteer/page/anywhereProfileConfirm/goVolunteerAnywhereProfileConfirm.html
+++ b/crossroads.net/app/go_volunteer/page/anywhereProfileConfirm/goVolunteerAnywhereProfileConfirm.html
@@ -1,0 +1,10 @@
+<div ng-if="$ctrl.viewReady">
+  <preloader full-screen='false' ng-show='!$ctrl.viewReady'> </preloader>
+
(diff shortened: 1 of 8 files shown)


crossroads.net/app/go_volunteer/page/anywhereProfileConfirm/goVolunteerAnywhereProfileConfirm.html
```diff
@@ -0,0 +1,9 @@
+<div ng-if="$ctrl.viewReady">
+  <preloader full-screen='false' ng-show='!$ctrl.viewReady'> </preloader>
```
Ganonside: will this preloader ever show up? it looks like if the preloader would be showing the div that it is contained in won't be.

Ganonside: I verified that it works but that isn't how I would have expected it to be written.

PR #3687, opened 2017-03-03
Feature/us6245 say hi email

ilpadre:


Gateway/crds-angular/Models/Finder/PinDto.cs
```diff
@@ -18,10 +18,10 @@ public enum PinType
 
     public class PinDto
     {
-        [JsonProperty("firstName")]
+        [JsonProperty("firstname")]
```
JoeKer1: I think these should be camel case or it will break other things

PR #3709, opened 2017-03-09
Check in test data

candrews-thrive:
Add user for testing CheckIn


hookom:
Don't we need the corresponding teardown script(s) to accompany these creates?

candrews-thrive:
The data will get deleted with TearDownUsers script

hookom:
Do our teardown scripts account for inserts like this one:
INSERT INTO [dbo].dp_user_roles 
 (User_ID,Role_ID,Domain_ID) VALUES
 (@userId,112      ,1        );

They may, I just can't remember if we delete from the dp_user_roles table.

candrews-thrive:
I believe so I copy that code from KathrynCradduckTest.sql . I look and there is nothing in the tear down folder for that script.

PR #3712, opened 2017-03-09
Feature/us7266 filter group connectors

silbermm:


Gateway/crds-angular.test/Services/GroupConnectorServiceTest.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using AutoMapper;
```
andycanterbury: Unused usings up in here

Gateway/crds-angular/packages.config
```diff
@@ -48,7 +48,6 @@
   <package id="SlowCheetah" version="2.5.14" targetFramework="net45" />
   <package id="Swashbuckle" version="5.2.1" targetFramework="net45" />
   <package id="Swashbuckle.Core" version="5.2.1" targetFramework="net45" />
-  <package id="System.Net.Http" version="4.3.0" targetFramework="net45" />
```
andycanterbury: Was this intentional? I feel like it was not.

PR #3724, opened 2017-03-13
DE2911: update email templates for batch processing

k8tied1:


k8tied1:
I did double check to ensure that the id's are the same in production. Feel free to double check!

PR #3739, opened 2017-03-14
DE2961 Fixed issue except one case

Ganonside:
If you are in multiple active grade groups you won't show up on the report
correctly.


Ganonside:
Could not run script. Please run for me.

PR #3755, opened 2017-03-17
US7176 paper contribution statements

chris-branch:


CI/SQL/20170316_171000_US7176-paper-statements-report-sp.sql
```diff
@@ -0,0 +1,233 @@
...
+		, Contact_Or_Household_ID
+		, Donor_ID
+		, Gender_ID
+		, Household_Position_ID
+		, Is_Recurring_Gift
+		, Section_Title
+		, Header_Title
+		, Section_Sort
+	)
+	SELECT TOP 100 PERCENT
```
k8tied1: What does this do? I looked around and it looks like this was a trick used with views in old version of SQL Server. Does it still do something relevant? If not, let's remove old MP relics
SELECT TOP 100 PERCENT

CI/Reports/CRDS ContributionStatement .rdl
```diff
@@ -0,0 +1,2021 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
k8tied1: Can we fix the name of the file to not have a space at the end?

CI/SQL/20170316_171000_US7176-paper-statements-report-sp.sql
```diff
@@ -0,0 +1,233 @@
...
+		, Section_Sort = CASE WHEN D.Payment_Type_ID = 6 THEN 2 ELSE 1 END
+	FROM Donations D
+		INNER JOIN #DONORS ON #DONORS.Donor_ID = D.Donor_ID
+		INNER JOIN #D ON #D.Donor_ID = D.Donor_ID
+		INNER JOIN Donors Do ON Do.Donor_ID = D.Donor_ID
+		INNER JOIN Contacts C ON C.Contact_ID = Do.Contact_ID
+		OUTER APPLY (SELECT Top 1 Contact_ID, Household_ID, Last_Name, Nickname, First_Name, Gender_ID FROM Contacts S WHERE Do.Statement_Type_ID = 2 AND S.Household_ID = C.Household_ID AND C.Contact_ID <> S.Contact_ID AND S.Household_Position_ID = 1 AND C.Household_Position_ID = 1) SP
+		LEFT OUTER JOIN Households H ON H.Household_ID = C.Household_ID
+		LEFT OUTER JOIN Addresses A ON A.Address_ID = H.Address_ID
+		INNER JOIN Donation_Distributions DD ON DD.Donation_ID = D.Donation_ID AND (@AddSoftCredit = 0 OR DD.Soft_Credit_Donor IS NULL)
```
k8tied1: It's a little confusing to have the @AddSoftCredit check here. We will include financial institutions with this logic. I think that this should just be DD.Soft_Credit_Donor is null. We will have to add in the soft credit donations with a different select statement anyway since we've already joined donation to donor (#D) on donor_id. For soft credit that join has to be donation to donation distribution to donor.

CI/SQL/20170316_171000_US7176-paper-statements-report-sp.sql
```diff
@@ -0,0 +1,233 @@
...
+			FROM Donations D
+			WHERE D.Donor_ID = Do.Donor_ID AND (Donation_Date >= @FromDate AND Donation_Date < @ToDate)
+		)
+		AND Do.Statement_Type_ID = 1		-- TODO: Individual only for now; remove this when adding Family
+	;
+
+	CREATE INDEX IX_D_DonorID ON #D(Donor_ID);
+
+	CREATE TABLE #DONORS (Donor_ID INT, Statement_ID VARCHAR(15));
+	INSERT INTO #DONORS SELECT DISTINCT Donor_ID, Statement_ID FROM #D;
```
k8tied1: I don't think that we actually need this yet since we aren't using it for anything (in the select below we could just as easily use statement_id from #D). Can we add a a note that it's for future use?

PR #3776, opened 2017-03-22
Defect/de2998 soft credit only donoations affect giving history

aiqingyuyan:


crossroads.net/app/giving_history/giving_history.controller.js
```diff
@@ -96,6 +97,11 @@
     };
 
     function setErrorState(error) {
+      // server error happens
+      if (error && error.status >= 500) {
+        vm.server_error = true;
```
k8tied1: I don't think that we can assume this. I think that there has to be changes on the server also to retun different errors for if donations==null and !donations.HasDonations
 if (donations == null ) || !donations.HasDonations)
  {
       return (RestHttpActionResult<ApiErrorDto>.WithStatus(HttpStatusCode.NotFound, new ApiErrorDto("No matching donations found")));
  }

PR #3787, opened 2017-03-24
DE3053: Dont check numbers if you don't have them

k8tied1:


k8tied1:
Joe K did code review

PR #3811, opened 2017-03-29
Family display name fix

lknair2015:


chris-branch:
Build failure looks like it's due to TeamCity weirdness (test that failed is related to Stripe, and this commit contains only .sql so completely unrelated).

PR #3820, opened 2017-03-31
Defect/de3098 full address

Doug-Shannon:
gets full address for places you are allowed to see it, and caches the result


Gateway/crds-angular/Services/GroupService.cs
```diff
@@ -501,7 +501,7 @@ public List<GroupDTO> GetGroupsByTypeForParticipant(string token, int participan
         /// <param name="groupTypeId"></param>
         /// <param name="groupId"></param>
         /// <returns></returns>
-        public List<GroupDTO> GetGroupsByTypeOrId(string token, int? participantId = null, int[] groupTypeIds = null, int? groupId = null)
+        public List<GroupDTO> GetGroupsByTypeOrId(string token, int? participantId = null, int[] groupTypeIds = null, int? groupId = null, bool? withParticipants = true, bool? withAttributes = true)
```
JoeKer1: This method's params are getting out of hand :(

PR #3843, opened 2017-04-05
US7707: improve sign-up-to-serve performance

aiqingyuyan:


chris-branch:
This was originally going to be a hotfix, but we decided to pull back and include it as part of the regular release cycle.

PR #3850, opened 2017-04-06
US7572: Run script nightly to detect contact email changes since yest…

k8tied1:
…erday

and update teh dp_user if we should and can


k8tied1:
1. Yes, will make the change to the script
2. it will, but i'm kind of ok with that. it will have the correct info for before and after and so it'll really just be logging that we detected it. And this is very very unlikely.
3. #Loginchanges will only pick them up if they don't match - c.Email_Address <> u.User_Name. So if we just changed user_name to match email_address they will match.  If we skip it due to duplicates, we won't create an audit record: the cursor uses the where clause OkToUpdate=1. 
4. That's both interesting and scary

PR #3852, opened 2017-04-07
US7762 - default AWS request size to config file

plachmann:


Gateway/crds-angular/Services/AwsCloudsearchService.cs
```diff
@@ -119,26 +120,26 @@ public AwsBoundingBox BuildBoundingBox(string upperleftlat , string upperleftlng
...
             }
-            else
-            {
-                searchRequest.Size = 31;
-            }
+            //else
+            //{
+                    //searchRequest.Size = _configurationWrapper.GetConfigIntValue("ConnectDefaultNumberOfPins");
+
+            //}
```
vchtelmakh: Should we remove the comments here?

PR #3861, opened 2017-04-11
Defect/de3151 redirect after login

richardbarker:


crossroads.net/core/login/login_controller.js
```diff
@@ -99,10 +101,20 @@
...
+                if (foundState.length > 0) {
+                  if (params === undefined) {
+                    $state.go(url);
+                  } else {
+                    $state.go(url, JSON.parse(params));
+                  }
                 } else {
-                  $state.go(url, JSON.parse(params));
+                  if (params === undefined) {
+                    $location.url(url);
```
mjvcallibrity: The `if (params === undefined)` check is happening here as well as up on line 107. Is it possible to extract this check out and place it before the `if (foundState.length > 0)` check?

PR #3862, opened 2017-04-11
DE3239: SU2S Front end back to asking for default (0) on init. Change

k8tied1:
backend to send minimum number of weeks until we find an opportunity (up
to 6). Also fix front end filter date range display

### DIFF
--- a/crossroads.net/app/my_serve/myserve.controller.js
+++ b/crossroads.net/app/my_serve/myserve.controller.js
@@ -73,8 +73,10 @@
     ////////////////////////////
 
     function activate() {
-      //vm.lastDate = formatDate(new Date(), 8);
-      vm.lastDate = formatDate(vm.groups[vm.groups.length - 1].day );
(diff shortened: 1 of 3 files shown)


crossroads.net/app/my_serve/myserve.controller.js
```diff
@@ -73,7 +73,8 @@
     ////////////////////////////
 
     function activate() {
-      vm.lastDate = formatDate(new Date(), 28);
+      //vm.lastDate = formatDate(new Date(), 8);
+      vm.lastDate = formatDate(vm.groups[vm.groups.length - 1].day );
```
chris-branch: Are we sure that vm.groups is not null and length > 0

k8tied1: Changes made. Also updated the server to bail on trimming down the list if it's empty or 1 item to begin with.

PR #3877, opened 2017-04-13
update how sanitizedDescription is set and use it in template

tatelucas:


crossroads.net/app/media/singleMedia.controller.js
```diff
@@ -56,7 +56,7 @@
     vm.showAudioDownloadLink = showAudioDownloadLink;
     vm.showProgramDownloadLink = showProgramDownloadLink;
     vm.shareUrl = $location.absUrl();
-    vm.sanitizedDescription = $sanitize(vm.media.description);
+    vm.sanitizedDescription = vm.media.description.replace(/<(?:.|\n)*?>/gm, '');
```
silbermm: test?

silbermm: well it looks like there are no tests for media at all... :cry:

PR #3911, opened 2017-04-25
US7389 Group Leader Confirmation

silbermm:
* added new method to set applicant as applied
* removed some un-needed attributes from MPHousehold
* updated config to included Applied GroupStatusId
* Tests!


Gateway/MinistryPlatform.Translation/Models/MpHousehold.cs
```diff
@@ -13,8 +13,7 @@ public class MpHousehold
 
         [JsonProperty(PropertyName = "Household_ID")]
         public int Household_ID { get; set; }
-
-        [JsonProperty(PropertyName = "Address_ID")]
```
silbermm: I added it in an earlier story (profile) and we don't use it anymore

PR #3941, opened 2017-04-28
Defect/de3392 error updating gathering

JoeKer1:


Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -462,6 +462,8 @@ public List<PinDto> GetMyGroupPins(string token, int[] groupTypeIds, int partici
             {
                 var pin = Mapper.Map<PinDto>(group);
                 pin.Gathering = group;
+                pin.Participant_ID = participantId;
+                pin.ShowOnMap = true;
```
sseissiger: Do we really need these for groups - guessing we do on the front end since you added them.  But I don't think we should.  At least showOnMap, that is only a person pin thing.  If a group comes back, then it should always show.

JoeKer1: participant_id is necessary. ShowOnMap is probably not, but if someone saved it for whatever reason it would go back to participant and then could get a little weird. I can take that one out if necessary

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -378,15 +385,26 @@ private static List<PinDto> ConvertFromAwsSearchResponse(SearchResponse response
...
                 {
+                    
                     pin.Gathering = new GroupDTO
                     {
                         GroupId = hit.Fields.ContainsKey("groupid") ? Convert.ToInt32(hit.Fields["groupid"].FirstOrDefault()) : 0,
                         GroupName = hit.Fields.ContainsKey("groupname") ? hit.Fields["groupname"].FirstOrDefault() : null,
                         GroupDescription = hit.Fields.ContainsKey("groupdescription") ? hit.Fields["groupdescription"].FirstOrDefault() : null,
                         PrimaryContactEmail = hit.Fields.ContainsKey("primarycontactemail") ? hit.Fields["primarycontactemail"].FirstOrDefault() : null,
-                        ContactId = hit.Fields.ContainsKey("countactid") ? Convert.ToInt32(hit.Fields["countactid"].FirstOrDefault()) : 0,
-                        Address = pin.Address
```
danrye: Why is Address missing now?

PR #3944, opened 2017-04-28
Defect/de3382 multi rendered headers crdsangular

richiebrees:
Fixes duplicate header and mobile issues in Legacy implementation of CRDS-SHARED-HEADER.

### DIFF
--- a/crossroads.net/package.json
+++ b/crossroads.net/package.json
@@ -108,7 +108,7 @@
     "api-check": "~7.5.5",
     "blueimp-load-image": "^1.14.0",
     "bootstrap": "^3.3.7",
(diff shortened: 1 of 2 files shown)


crossroads.net/package.json
```diff
@@ -108,7 +108,7 @@
     "api-check": "~7.5.5",
     "blueimp-load-image": "^1.14.0",
     "bootstrap": "^3.3.7",
-    "crds-shared-header": "0.2.10",
+    "crds-shared-header": "^0.2.11",
```
tcmacdonald: Recommend loosing the caret here.

richiebrees: Good call!

PR #3948, opened 2017-04-29
store longitude in aws too...

JoeKer1:


jpc11623:
I looked at this and am okay with the code. Joe will merge when it has a good build.

PR #3955, opened 2017-05-02
DE3447: Left join on pledges to dd instead of inner join to include

k8tied1:
pledges with no donations

### DIFF
--- a/CI/SQL/20170501_110100_DE3447-Include0PercentPledges.sql
+++ b/CI/SQL/20170501_110100_DE3447-Include0PercentPledges.sql
@@ -1,0 +1,410 @@
+USE [MinistryPlatform]
+GO
+
+/****** Object:  StoredProcedure [dbo].[report_CRDS_Donor_Contribution_Statement]    Script Date: 4/19/2017 12:41:11 PM ******/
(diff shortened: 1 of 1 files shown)


CI/SQL/20170501_110100_DE3447-Include0PercentPledges.sql
```diff
@@ -0,0 +1,410 @@
...
+	--If they selected none in the list we aren't going to give them any even if they selected other ones
+	IF NOT EXISTS (Select 1 from #Campaigns where campaign_id = 0)
+	BEGIN
+
+	  --We can't join on donors yet because we want all donations for the pledge regardless of the donor
+	
+	  INSERT INTO #PledgeTotals 
+	  SELECT p.Pledge_ID, p.Donor_ID, p.Total_Pledge, SUM (dd.amount),p.campaign_name
+	  FROM #Pledges p
+	  LEFT JOIN Donation_Distributions dd ON dd.Pledge_ID =p.Pledge_ID
```
chris-branch: Seems like we want only "Deposited" donations here, so probably need to left join with donations and filter by Donation_Status_ID = 2

PR #3967, opened 2017-05-04
Defect/de3443 rebase

plachmann:


Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -352,6 +360,39 @@ public List<PinDto> GetPinsInBoundingBox(GeoCoordinate originCoords, string addr
...
+        private string GetPinTitle(PinDto pin)
+        {
+            string jsonData="";
+            switch (pin.PinType)
+            {
+                case PinType.SITE:
+                    jsonData = $"{{ 'siteName': '{pin.SiteName}','isHost':  false,'isMe': false,'pinType': {(int)pin.PinType}}}";
+                    break;
+                case PinType.GATHERING:
+                    jsonData = $"{{ 'firstName': '{pin.FirstName}', 'lastInitial': '{pin.LastName[0]}','isHost':  true,'isMe': false,'pinType': {(int)pin.PinType}}}";
```
chris-branch: Is pin.LastName guaranteed to be non-null and length >= 1 ?  If not, this will generate an exception.

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -352,6 +360,39 @@ public List<PinDto> GetPinsInBoundingBox(GeoCoordinate originCoords, string addr
...
+                    break;
+                case PinType.PERSON:
+                    jsonData = $"{{ 'firstName': '{pin.FirstName}', 'lastInitial': '{pin.LastName[0]}','isHost':  false,'isMe': false,'pinType': {(int)pin.PinType}}}";
```
chris-branch: Same as above.  Is pin.LastName guaranteed to be non-null and length >= 1 ?  If not, this will generate an exception.

PR #3970, opened 2017-05-05
DEXXXX Long Build Times

tcmacdonald:
This reduces the time spent in `gulp build` from 4.64 min to 1.42 min.

### DIFF
--- a/crossroads.net/app/index.html
+++ b/crossroads.net/app/index.html
@@ -94,6 +94,11 @@
     <script src="https://use.typekit.net/ccb3vpa.js"></script>
     <script>try{Typekit.load({ async: true });}catch(e){}</script>
 
+
+    <!-- build:legacycss -->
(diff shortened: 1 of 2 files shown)


crossroads.net/app/index.html
```diff
@@ -94,6 +94,11 @@
     <script src="https://use.typekit.net/ccb3vpa.js"></script>
     <script>try{Typekit.load({ async: true });}catch(e){}</script>
 
+
+    <!-- build:legacycss -->
+    <link href="/assets/legacy.css" rel="stylesheet" />
```
charlie-retzler: Would we also need this in crds-maestro template for legacy?

tcmacdonald: @charlie-retzler `crds-maestro` is already consuming legacy.css. This update simply ensures we don't waste time building those styles twice.

PR #3991, opened 2017-05-19
Feature/us8234 group search in finder

vchtelmakh:


CI/SQL/20170518_131422_US8234-small-groups-in-finder.sql
```diff
@@ -0,0 +1,238 @@
...
+		EXEC(@DynamicQuery)
+
+	END
+	END
+
+END
+
+
+GO
+
```
JoeKer1: Can you show me what changed from what was out there? It's hard to tell. Also make sure old group tool still works.

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -231,25 +238,25 @@ private void GatheringValidityCheck(int contactId, AddressDTO address)
```
JoeKer1: Right now this method goes and gets all the groups and that seems wasteful. Can we create a new method that gets a list of groups filtered by this data already?

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -348,42 +355,61 @@ public int GetLeaderParticipantIdFromGroup(int groupId)
...
+            int id = _anywhereGroupType; 
+
+            if (finderType.Equals(_finderConnect))
             {
-                pin.Title = GetPinTitle(pin);
-                pin.IconUrl = GetPinUrl(pin.PinType);
-                //calculate proximity for all pins to origin
-                if (pin.Address.Latitude == null) continue;
-                if (pin.Address.Longitude != null) pin.Proximity = GetProximity(originCoords, new GeoCoordinate(pin.Address.Latitude.Value, pin.Address.Longitude.Value));
+                var cloudReturn = _awsCloudsearchService.SearchConnectAwsCloudsearch("matchall",
```
JoeKer1: nitpicky but we could encapsulate these in config

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -348,42 +355,61 @@ public int GetLeaderParticipantIdFromGroup(int groupId)
...
+                                                                                     "_all_fields",
+                                                                                     _configurationWrapper.GetConfigIntValue("ConnectDefaultNumberOfPins"),
```
JoeKer1: Not sure if "ConnectDefaultNumberOfPins is used anywhere else. Probably should just set a constant in the constructor so we can use it where ever.

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -348,42 +355,61 @@ public int GetLeaderParticipantIdFromGroup(int groupId)
...
             }
 
+            this.AddPinMetaData(pins, originCoords);
             return pins;
         }
 
         private string GetPinTitle(PinDto pin)
         {
-            string jsonData="";
+            string jsonData = "";
```
JoeKer1: Why are we doing this in json? Should this be in a class and then we let the serializer do ti?

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -436,27 +464,27 @@ private List<PinDto> ConvertFromAwsSearchResponse(SearchResponse response)
...
-                    
+
                     pin.Gathering = new GroupDTO
                     {
                         GroupId = hit.Fields.ContainsKey("groupid") ? Convert.ToInt32(hit.Fields["groupid"].FirstOrDefault()) : 0,
                         GroupName = hit.Fields.ContainsKey("groupname") ? hit.Fields["groupname"].FirstOrDefault() : null,
                         GroupDescription = hit.Fields.ContainsKey("groupdescription") ? hit.Fields["groupdescription"].FirstOrDefault() : null,
                         PrimaryContactEmail = hit.Fields.ContainsKey("primarycontactemail") ? hit.Fields["primarycontactemail"].FirstOrDefault() : null,
-                        Address = pin.Address, 
+                        Address = pin.Address,
```
JoeKer1: Should this be pin.Gathering.Address?

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -466,43 +494,43 @@ private List<PinDto> ConvertFromAwsSearchResponse(SearchResponse response)
...
 
         private static double Proximity(double lat1, double lon1, double lat2, double lon2)
         {
             var theta = lon1 - lon2;
-            var dist = Math.Sin(Deg2Rad(lat1)) * Math.Sin(Deg2Rad(lat2)) + Math.Cos(Deg2Rad(lat1)) * Math.Cos(Deg2Rad(lat2)) * Math.Cos(Deg2Rad(theta));
+            var dist = Math.Sin(Deg2Rad(lat1))*Math.Sin(Deg2Rad(lat2)) + Math.Cos(Deg2Rad(lat1))*Math.Cos(Deg2Rad(lat2))*Math.Cos(Deg2Rad(theta));
             dist = Math.Acos(dist);
             dist = Rad2Deg(dist);
-            dist = dist * 60 * 1.1515;
+            dist = dist*60*1.1515;
```
JoeKer1: Magic, move these to configs or const?

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -763,5 +778,73 @@ private void SendGatheringInviteResponseEmail(bool inviteAccepted, PinDto host,
...
+
+        private List<PinDto> TransformGroupDtoToPinDto(List<GroupDTO> groupDTOs, string finderType)
+        {
+            var pins = new List<PinDto>();
+
+            if (finderType.Equals(_finderConnect))
+            {
+                foreach (var group in groupDTOs)
+                {
+                    var pin = Mapper.Map<PinDto>(group);
```
JoeKer1: If we take care of the TODO we might be able to setup automapper to do this for us. Probably still the same just less code.  Maybe, the pin type might not let us do that.

Gateway/crds-angular/Services/FinderService.cs
JoeKer1: Could even just return the count directly from the rest api.

CI/SQL/20170518_131422_US8234-small-groups-in-finder.sql
vchtelmakh: I just tested search on int.crossroads.net/groups - it works.

Gateway/crds-angular/Services/FinderService.cs
vchtelmakh: I changed them to constants, but they are pretty magical anyway. I have to assume someone just copied formulas from http://stackoverflow.com/questions/389211/geospatial-coordinates-and-distance-in-kilometers

Gateway/crds-angular/Services/FinderService.cs
vchtelmakh: Not as far as I can tell - data comes back flat from AWS, we want the address on the top level.

Gateway/crds-angular/Services/FinderService.cs
vchtelmakh: We need this as JSON on the front end to draw the labels. Its convoluted and dirty, but in short, we have a title attribute generated by NG2 Goog maps that we need to stuff label data into. We can't pass an object to the title, so we pass a JSON string. This causes parsing errors if group names have apostrophes, etc. We have a story to handle this.

CI/SQL/20170518_131422_US8234-small-groups-in-finder.sql
JoeKer1: 👍

PR #3992, opened 2017-05-19
US7537 shared header

tcmacdonald:


crossroads.net/app/index.html
```diff
@@ -145,6 +145,7 @@
 </body>
 
 <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
+<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.js"></script>
 <script type="text/javascript" src="/assets/load-image.all.min.js" defer> </script>
```
richardbarker: Is this so that we can run legacy stand-alone?

tcmacdonald: @richardbarker Thats correct.

PR #3996, opened 2017-05-22
Defect/de3460 database refresh fix

aiqingyuyan:


CI/DatabaseRestoreFromBlob.ps1
```diff
@@ -55,6 +61,33 @@ $connection = New-Object System.Data.SqlClient.SqlConnection;
...
+    32 = "DeceasedPerson";
+    375 = "ResetUser";
+    601 = "CustomFormEditor";
+    602 = "SendGrid";
+}
+
+$sqlInsertValues = "";
+
+foreach ($key in @($toolsToBeUpdated.Keys)) {
+    $toolsToBeUpdated[$key] = $dpToolUriPrefix + $DpToolBaseUpdateUriToken + $toolsToBeUpdated[$key] + "/"
```
charlie-retzler: @aiqingyuyan Would it be possible to change this to just send in the value for $dpToolUriPrefix + $DpToolBaseUpdateUriToken would evaluate too instead of sending in the 3 separate fields ($Branch, $dbToolUriPrefix and $DpToolBaseUpdateUriToken)?

CI/DatabaseRestoreFromBlob.ps1
```diff
@@ -55,6 +61,33 @@ $connection = New-Object System.Data.SqlClient.SqlConnection;
...
+    15 = "ContactLog";
+    32 = "DeceasedPerson";
+    375 = "ResetUser";
+    601 = "CustomFormEditor";
+    602 = "SendGrid";
+}
+
+$sqlInsertValues = "";
+
+foreach ($key in @($toolsToBeUpdated.Keys)) {
```
chris-branch: We could simplify the logic by just doing a global find/replace on the Launch_Page column in dp_Tools.  This would also handle any additional tools that TM adds in the future without requiring an update to this script.  Something like this (but with parameters in place of hard-coded values):

UPDATE dp_Tools SET Launch_Page = REPLACE(Launch_Page, 'crossroads.ministryplatform.com', 'crossroads-demo.ministryplatform.com')

PR #3998, opened 2017-05-22
Feature/us7391 reference notification

silbermm:


Gateway/crds-angular/Services/GroupLeaderService.cs
```diff
@@ -179,6 +178,66 @@ public IObservable<int> SaveSpiritualGrowth(SpiritualGrowthDTO spiritualGrowth)
...
+        public IObservable<Dictionary<string, object>> GetReferenceData(int contactId)
+        {
+            var formId = _configWrapper.GetConfigIntValue("GroupLeaderFormId");
+            var formFieldId = _configWrapper.GetConfigIntValue("GroupLeaderFormReferenceContact");
+
+            return Observable.Return<MpParticipant>(_participantRepository.GetParticipant(contactId)).Zip(
+                Observable.Return<MpMyContact>(_contactRepository.GetContactById(contactId)),
+                Observable.Return<string>(_formSubmissionRepository.GetFormResponseAnswer(formId, contactId, formFieldId, null)),
+                (participant, contact, answer) => new Dictionary<string, object>
+                {
```
andycanterbury: An observable of an observable of an observable??

Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -210,6 +211,9 @@
   <add key="GroupLeaderHuddleFieldId" value="1520"/>
   <add key="GroupLeaderStudentFieldId" value="1521"/>
   <add key="GroupLeaderConfirmationTemplate" value="2015"/>
+  <add key="GroupLeaderReferenceEmailTemplate" value="2018"/>
+  <add key="GroupsContactId" value="7675411"/>
```
andycanterbury: I'm not sure what we do with other emails, but could we use the contact and email from the template so that they are user changeable instead of putting them in the config file and requiring a deploy to change them?

Gateway/crds-angular/MinistryPlatform.config
```diff
@@ -320,6 +324,7 @@
   <add key="GroupAgeRangeAttributeTypeId" value="91"/>
   <add key="GroupLeaderRoleId" value="22"/>
   <add key="MyCurrentGroupParticipationPageId" value="563"/>
+  <add key="GroupsEmailAddress" value="groups@crossroads.net" />
```
andycanterbury: See above comment

Gateway/crds-angular/Services/GroupLeaderService.cs
silbermm: yes, Observable.zip takes observables and runs them async and the last function takes all of the return values from the observables as an argument. see [the docs](http://reactivex.io/documentation/operators/zip.html)

Gateway/crds-angular/MinistryPlatform.config
silbermm: Your right, that is probably the right thing to do, but would require yet another call to the server to get that info.

I'll look into it

PR #4003, opened 2017-05-23
US7966: Resolve deadlocks with GetByAuthenticationToken()

chris-branch:


Gateway/MinistryPlatform.Translation/Repositories/UserRepository.cs
```diff
@@ -50,10 +54,13 @@ public MpUser GetUserByRecordId(int recordId)
...
-            var contactId = _ministryPlatformService.GetContactInfo(authToken).ContactId;
+            var userId = _ministryPlatformService.GetContactInfo(authToken).UserId;
 
-            var searchString = string.Format(",\"{0}\"", contactId);
-            return (GetUser(searchString));
+            string search = $"User_ID = {userId}";
+            string columns = "User_ID, User_Name, User_Email, User_GUID, Can_Impersonate";
+            var userList = _ministryPlatformRest.UsingAuthenticationToken(ApiLogin()).SearchTable<MpUser>("dp_Users", search, columns);
+
+            return userList.First();
```
IngageKrizJ: I think (not 100% sure though) that the previous implementation calling GetUser() would return null if the user was not found, whereas this will end up throwing an exception (calling First() on an empty list).

chris-branch: Good catch.  The previous version does return null instead of throwing (though couldn't find any callers that check for null, so the end result will be the same).  My goal was to not change behavior, so I'll update the code to use FirstOrDefault() instead of First().

PR #4034, opened 2017-05-25
Feature/us7245 event setup report

k8tied1:


CI/SQL/20170525_133500_US7245-event-setup_report-update-sp.sql
```diff
@@ -0,0 +1,328 @@
...
+
+	   -- try this to make sure building field is populated for something besides Oakley
+	   SELECT @BuildingID = building_id
+	   FROM Buildings
+	   WHERE location_id = @BuildingID
+
+      SELECT E.event_id
+        INTO #e
+        FROM events E
+		JOIN event_participants p on E.event_id = p.event_id
```
chris-branch: Should the join to event_participants be deleted?  The query doesn't appear to reference any columns from this table, but if there are 0 rows it will prevent the event from showing up in the report.

k8tied1: Addressed. Thanks!

PR #4045, opened 2017-05-31
US8121 - Clean up End_Date for Group Participant Attributes

chris-branch:
* New trigger that will sync End_Date in Group_Participant_Attributes
* New index on Group_Participant_Attributes to ensure the trigger is efficient
* One-time query to clean up existing data

### DIFF
--- a/CI/SQL/20170531_095800_US8121-end-date-group-participant-attributes.sql
+++ b/CI/SQL/20170531_095800_US8121-end-date-group-participant-attributes.sql
@@ -1,0 +1,26 @@
+﻿USE [MinistryPlatform]
+GO
+
+CREATE INDEX IX_Group_Participant_Attributes__GroupParticipantID ON Group_Participant_Attributes(Group_Participant_ID);
+GO
+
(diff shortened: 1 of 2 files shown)


CI/SQL/20170531_095800_US8121-end-date-group-participant-attributes.sql
```diff
@@ -0,0 +1,26 @@
...
+    SET NOCOUNT ON;
+
+	-- When Group_Participants.End_Date is modified, apply the same change to Group_Participant_Attributes.End_Date
+	IF UPDATE(End_Date)
+	BEGIN
+		UPDATE gpa
+		SET gpa.End_Date = gp.End_Date
+		FROM
+			INSERTED i
+			INNER JOIN Group_Participants gp ON gp.Group_Participant_ID = i.Group_Participant_ID
```
k8tied1: Do you have to join to Group_Participants here? I think that inserted should have the value of end_date so you should be able to set gpa.end_date = i.end_date

PR #4055, opened 2017-06-01
Fixed county field's name value

s-henney:
Both the County and Country fields had name values of "country". Renamed County's name to "county".


s-henney:
renaming branch

PR #4064, opened 2017-06-01
Feature/us8238 accept deny request sm group

sseissiger:


Gateway/crds-angular/Services/GroupToolService.cs
```diff
@@ -118,13 +120,16 @@ IFinderRepository finderRepository
             _baseUrl = configurationWrapper.GetConfigValue("BaseURL");
             _addressMatrixSearchDepth = configurationWrapper.GetConfigIntValue("AddressMatrixSearchDepth");
             
-            _smallGroupTypeId = configurationWrapper.GetConfigIntValue("SmallGroupTypeId");
+            _smallGroupTypeId = configurationWrapper.GetConfigIntValue("SmallGroupTypeId" +
+                                                                       "");
```
JoeKer1: I don't think we need the + ""

PR #4072, opened 2017-06-02
Feature/us8458 invoice confirmation email

timmyg:
build: https://goo.gl/xcAxcc


crossroads.net/app/invoices/invoice-confirmation-redirecting.html
```diff
@@ -0,0 +1,6 @@
+<div>
```
jpc11623: Should this be a content block? I feel that Finance will have some concerns about the language

PR #4083, opened 2017-06-05
US8122 - FinishSendMessageFromDonor Errors

aiqingyuyan:


k8tied1:
Please update the AC on the story to correctly reflect the changes in requirements. There's nothing about trips - if this is by design, that's fine but let's keep the story accurate for testing and for progeny.

PR #4104, opened 2017-06-08
US8613 - Improve stripe error handling

chris-branch:


Gateway/crds-angular/Services/StripePaymentProcessorService .cs
```diff
@@ -342,22 +364,34 @@ public List<StripeCharge> GetChargesForTransfer(string transferId)
             } while (nextPage.HasMore);
 
             //get the metadata for all of the charges
-            foreach (var charge in charges)
+            for (int i = charges.Count - 1; i >= 0; --i)
```
IngageKrizJ: This seems like we are going to miss the last charge - I could be wrong though.  I think `--i ` is going to decrement right away (before the first loop iteration), so is it starting at charges.Count - 2?

Gateway/crds-angular/Services/StripePaymentProcessorService .cs
```diff
@@ -81,33 +85,51 @@ private PaymentProcessorException AddGlobalErrorMessage(PaymentProcessorExceptio
...
             }
             else if ("bank_account".Equals(e.Param))
             {
                 if ("invalid_request_error".Equals(e.Type))
                 {
-                    e.GlobalMessage = _contentBlockService["paymentMethodDeclined"];
+                    e.GlobalMessage = GetContentBlock("paymentMethodDeclined");
                 }
             }
             else
```
chris-branch: Yes, that makes sense.  I'll make the change.

Gateway/crds-angular/Services/StripePaymentProcessorService .cs
chris-branch: The decrement runs at the conclusion of each loop iteration, so the first index will be Count - 1 and the last index will be 0, so I think we should be good.

aiqingyuyan:
Looks good to me

PR #4108, opened 2017-06-09
DE3649 - Sign Up To Serve Icons

seancdavis:
- Fix color on checkmark icon
- Fix popovers


seancdavis:
@tcmacdonald came up with a cleaner solution to this problem. Foregoing this for his PR.

PR #4135, opened 2017-06-19
de2608 disable $log.debug on prod

aiqingyuyan:


crossroads.net/app/common/profile/household/profileHousehold.directive.js
```diff
@@ -5,7 +5,10 @@
   ProfileHouseholdDirective.$inject = ['$log'];
 
   function ProfileHouseholdDirective($log) {
-    $log.debug('ProfileHouseholdDirective');
+    // don't log anything in prod environment
```
chris-branch: We should be able to disable logging in one place at a global level instead of modifying each call to debug().  I think we just want to call $logProvider.debugEnabled(false) for PROD.

PR #4140, opened 2017-06-20
DE3675 - Exclude cancelled events from the sign-up page.

chris-branch:


Gateway/MinistryPlatform.Translation/Models/MpEvent.cs
```diff
@@ -41,6 +41,7 @@ public class MpEvent
 
         private IList<int> participants = new List<int>();
 
+        [JsonProperty(PropertyName = "Congregation_Name")]
```
k8tied1: Did you add this on purpose? It doesn't seem to fit with everything else...

Gateway/MinistryPlatform.Translation/Repositories/GroupRepository.cs
```diff
@@ -569,23 +569,25 @@ private List<MpGroupParticipant> LoadGroupParticipants(int groupId, string token
             return groupParticipants;
         }
 
-        public IList<MpEvent> getAllEventsForGroup(int groupId)
+        public IList<MpEvent> getAllEventsForGroup(int groupId, DateTime? minEndDate = null, bool includeCancelledEvents = false)
```
k8tied1: I didn't see anywhere that you've updated where you are calling this to pass in the parameters. Did I miss it or is it not checked in?

Gateway/MinistryPlatform.Translation/Models/MpEvent.cs
chris-branch: Yes, on purpose.  It's using the REST API now and relying on the automatic mapping from json-to-MpEvent.  The Congregation field was already there but not decorated with the [JsonProperty] attribute, so I had to add that attribute to get the automatic mapping for that field.

Gateway/MinistryPlatform.Translation/Repositories/GroupRepository.cs
chris-branch: Both of the new parameters have default values, so the callers don't need to be updated as long as the defaults are fine (which is true for now).  Eventually, I think we can reduce the amount of data we pull by specifying a minimum date (and we will actually need to do that for DE2903 which is still on the board in ToDo), but for now I left the existing behavior as is.

PR #4141, opened 2017-06-21
US8395 - app participant directly to group

plachmann:


Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -420,6 +423,20 @@ public List<PinDto> GetPinsInBoundingBox(GeoCoordinate originCoords, string addr
...
 
+        public void AddUserDirectlyToGroup(User user, int groupid)
+        {
+            //check to see if user exists in MP.
+            var contactId = _contactRepository.GetContactIdByEmail(user.email);
+            if (contactId == 0)
+            {
+                user.password = System.Web.Security.Membership.GeneratePassword(25, 10);
+                _accountService.RegisterPersonWithoutUserAccount(user);
+                contactId = _contactRepository.GetContactIdByEmail(user.email);
```
sseissiger: would it make more sense/possible to return the contactId from the above call to register the user account

PR #4173, opened 2017-06-28
iDE3793 - Mobile iOS Scroll

seancdavis:
Fix scrolling issue on iOS

---

Logged in or out on iOS devices.
The main menu, as well as the menu under My Profile the page only partially loads (you can't scroll) unless you refresh. 

Notes from Tate:
it looks like at some point the class 'noscroll' is getting added to the body. When that happens, the page content is cut off and you can't scroll down the page. From my testing, it seems to show up when you go to the login form. It doesn't matter if you log in or not from that point.

Looking at the legacy.scss file, the .noscroll stuff is right in the file

---

(The difficulty in testing this is that it wasn't previously broken in int.)


seancdavis:
Closing this because we're going to use a hotfix to get this change into release and will merge back into development.

PR #4214, opened 2017-07-14
US8988 - Travel Information Update UI

tallented:
http://localhost:3000/trips/mytrips/update

image

Mocked or stubbed most of the controller functions and frequent flyer data

### DIFF
--- a/crossroads.net/app/trips/travelInformation/travelInformation.controller.js
+++ b/crossroads.net/app/trips/travelInformation/travelInformation.controller.js
@@ -95,6 +95,20 @@ export default class TravelInformationController {
         "categoryId": null,
         "categoryDescription": null,
         "attributeTypeId": null
+      },
+      {
+        "attributeId": 5000,
+        "name": "American Airlines",
+        "description": null,
+        "selected": false,
+        "startDate": "0001-01-01T00:00:00",
+        "endDate": null,
+        "notes": null,
+        "sortOrder": 0,
+        "category": null,
+        "categoryId": null,
+        "categoryDescription": null,
+        "attributeTypeId": null
       }
     ];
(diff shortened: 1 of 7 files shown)


crossroads.net/app/trips/travelInformation/travelInformation.controller.js
```diff
@@ -0,0 +1,170 @@
...
+      }
+    ];
+
+    this.now = new Date();
+    this.initDate = new Date(this.now.getFullYear(), this.now.getMonth(), this.now.getDate());
+    this.maxPassportExpireDate = new Date(this.now.getFullYear() + 150, this.now.getMonth(), this.now.getDate());
+    this.minPassportExpireDate = new Date(this.now.getFullYear(), this.now.getMonth(), this.now.getDate());
+  }
+
+  showFrequentFlyer(airline) {
```
andycanterbury: This page should show all the airlines, because it is not trip specific. I think we can get rid of this method and the associated ones.

tallented: Ok, I'll remove the trip-specific checks

crossroads.net/app/trips/trips.routes.js
```diff
@@ -131,6 +131,38 @@
...
+            var cid = $cookies.get('userId');
+            if ($stateParams.contactId) {
+              cid = $stateParams.contactId;
+            }
+
+            return Profile.Person.get({ contactId: cid }).$promise;
+          },
+
+          Trip: 'Trip',
+          MyTrips: function(Trip) {
```
silbermm: do we really need this when updating travel information? I'd guess no.

PR #4223, opened 2017-07-17
US9111: Make SendMail faster

k8tied1:


Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -161,6 +161,25 @@ public string GetContactEmail(int contactId)
             }
         }
 
+        public MpContact GetEmailFromDonorId(int donorId)
+        {
+            MpContact contact = new MpContact();
+            try
+            {
+                string searchFilter = $"Donor_Record={donorId}";
+                var foundEmails = _ministryPlatformRest.UsingAuthenticationToken(base.ApiLogin()).Search<MpContact>(searchFilter, "Contact_ID, Donor_Record, Email_Address");
```
chris-branch: Maybe a little counter-intuitive, but ... we would likely get better performance if this query could search Donors for a matching Donor_ID, then join to Contacts to get the email address.  Currently, it's searching Contacts.Donor_Record which does not have an index, so it's going to do a full table scan on Contacts.

k8tied1: It's 293 milliseconds right now. Your change only improves to 260 ms

PR #4233, opened 2017-07-19
Feature/us8162 update travel info

silbermm:


crossroads.net/app/trips/travelInformation/travelInformation.controller.js
```diff
@@ -24,113 +27,33 @@ export default class TravelInformationController {
...
+    }
+
+    this.attributeTypeService.AttributeTypes().get({ id: attributeType }, (data) => {
+      this.frequentFlyers = data.attributes.map((ff) => {
+        const exists = this.frequentFlyerValue(ff.attributeId);
+        if (exists) {
+          const newFF = Object.assign({}, ff, { selected: true, notes: exists, startDate: new Date() });
+          return newFF;
+        }
+        return ff;
```
Ganonside: the return could be reduced to a ternary

PR #4269, opened 2017-07-27
DE3903 - Profile Menu

seancdavis:
Steps to recreate:
On mobile, sign in, 
open the profile
navigate to any of the menu items under profile
the first time - profile menu collapses, but when you open any menu item (same or different) again profile stays expended.
There is video recording of all those steps on slack on Ample channel sent by Jason on 7/17/2017
and you can access this video here:
https://drive.google.com/drive/u/0/folders/0B0E7EftymfMHN3EwU3hYcjVkWnc?ths=true (the name of the video profile expanded)

### DIFF
--- a/crossroads.net/core/core.run.js
+++ b/crossroads.net/core/core.run.js
@@ -36,6 +36,8 @@
     function setupHeader() {
       svg4everybody();
       $('html, body').removeClass('noscroll');
+      $('.collapse.in').removeClass('in');
+      $('.modal-backdrop.fade.in').remove();
       // header options
       var options = {
         el: '[data-header]',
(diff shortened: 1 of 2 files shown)


crossroads.net/core/core.run.js
```diff
@@ -36,6 +36,8 @@
     function setupHeader() {
       svg4everybody();
       $('html, body').removeClass('noscroll');
+      $('.collapse.in').removeClass('in');
```
seancdavis: I was thinking I should probably scope this to the header, but because it was on state change I wasn't super worried. Thoughts?

tcmacdonald: Seems like a reasonable assumption. It doesn't appear to negatively impact the desktop menu so lets roll w/ it. Thanks!

PR #4273, opened 2017-07-31
Feature/us9163 events phase i general

ilpadre:


Gateway/crds-angular/Controllers/API/LoginController.cs
```diff
@@ -6,13 +6,15 @@
...
 using crds_angular.Services;
+using crds_angular.Services.Analytics;
 using crds_angular.Services.Interfaces;
 using MinistryPlatform.Translation.Models.DTO;
 using Crossroads.ApiVersioning;
 using Crossroads.ClientApiKeys;
 using Crossroads.Web.Common.Security;
 using MinistryPlatform.Translation.Repositories.Interfaces;
 
+
```
Doug-Shannon: ahem

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.Ajax.Utilities;
+using MinistryPlatform.Translation.PlatformService;
+using Segment;
+using Segment.Model;
+
+namespace crds_angular.Services.Analytics
+{
+    public class AnalyticsAstronomer
+    {
+        private string ASTRONOMER_APPLICATION_ID = "sZQew9P77arhJe3Qx";
```
Doug-Shannon: this needs to be in an environment variable.  This is just pointing to int.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public  void Track(string userId, string eventName, EventProperties props)
+        {
+            Properties segProps = mapProps(props);
+            Segment.Analytics.Initialize(ASTRONOMER_APPLICATION_ID, new Config().SetAsync(false));
```
Doug-Shannon: is this supposed to init every time we track, or is it supposed to init on bootstrap of the application?

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
```diff
@@ -0,0 +1,36 @@
...
+            Segment.Analytics.Client.Track(userId, eventName, segProps);
+            Segment.Analytics.Client.Flush();
```
Doug-Shannon: I know this was the way you were able to get events to show up immediately, but we may want to discuss this with Architecture to determine a better auto flush rate instead of sending with every call

Gateway/crds-angular/Services/Analytics/AnalyticsService.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace crds_angular.Services.Analytics
+{
+    public class AnalyticsService: IAnalyticsService
+    {
+        public void Track(string userId, string eventName)
+        {
+            EventProperties props = new EventProperties();
+            props.Add("Source", "CrossroadsNet");
+            var astronomer = new AnalyticsAstronomer();
```
Doug-Shannon: should we do this on init instead of every call?

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
Doug-Shannon: also, what is setAsync.. should that be true? we dont want to hold up calls on this.

Gateway/crds-angular/Services/Analytics/AnalyticsService.cs
ilpadre: It may be true for Source, but if there are other events that require additional properties, we would need to add them. The caller could be responsible for assembling properties, and we have an overloaded Track method that includes the properties in the signature. If these are unfounded concerns, then we can move them to init.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
ilpadre: SetAsync on Config just determines which thread flush works on. If SetAsync is false, flush will empty the buffer immediately. True will empty it on another thread. It depends on whether we want to have events show up immediately in Astronomer/MixPanel or if a delay is acceptable. It might be better to take a delay in the event showing up rather than hold up the method.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
ilpadre: SetAsync on Config just determines which thread flush works on. If SetAsync is false, flush will empty the buffer immediately. True will empty it on another thread. It depends on whether we want to have events show up immediately in Astronomer/MixPanel or if a delay is acceptable. It might be better to take a delay in the event showing up rather than hold up the method.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
ilpadre: Ahem.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
Doug-Shannon: I'm not sure what the right answer is for this.  I feel like it should be async, but I may not fully understand the implications.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
Doug-Shannon: is this supposed to init every time we track, or is it supposed to init on bootstrap of the application?

Gateway/crds-angular/Services/Analytics/AnalyticsService.cs
Doug-Shannon: I think we should put this in init until it is no longer feasible.  Its just weird to init the library over and over for each call.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
ilpadre: I'm pretty sure we can do it once at startup. I'll make the change and make sure everything is working ok.

Gateway/crds-angular/Services/Analytics/AnalyticsAstronomer.cs
ilpadre: I'll try it with the async true and see how that affects the performance. I don't know how it will act under heavy load. I'm sure the data team wants their events right away, but I'm also sure everyone else wants the methods to return quickly. The app is already slow.

Gateway/crds-angular/Services/Analytics/AnalyticsService.cs
ilpadre: Agreed on initializing the library. The properties are different for each event, or they can be, so that  can't be done at initialization time. The options there are, is it done by the caller or in the Track method. Since the source will always be CrossroadsNet from the server side, I don't see a point in making each caller create a properties object. If the caller needs to add additional properties, that's where the overloaded method would come in.

Gateway/crds-angular.test/controllers/UserControllerTest.cs
```diff
@@ -50,10 +53,13 @@ public void ShouldRegisterNewUser()
         {
             var user = new User();
             _accountService.Setup(mocked => mocked.RegisterPerson(user)).Returns(user);
+            //_mockAnalyticService.Setup(mocked => mocked.Track("test@email.com", "SignedUp"));
```
Doug-Shannon: Should we just delete this?  Why is it commented out

Gateway/crds-angular/Controllers/API/LoginController.cs
Doug-Shannon: double blank space `ahem`

Gateway/crds-angular/Services/Analytics/AnalyticsService.cs
Doug-Shannon: in the constructor we should `var astronomer = new AnalyticsAstronomer()` then just use astonomer where we need it, instead of creating a new one every time.

PR #4295, opened 2017-08-07
US9412: Add medical conditions column to report

k8tied1:


CI/SQL/20170807_111400_US9412-add-medicalconditios-to-sp-report_CRDS_Trip_Export.sql
```diff
@@ -0,0 +1,602 @@
...
+
+   (SELECT TOP 1 notes
+	   FROM vw_crds_contact_attributes
+	   WHERE contact_id = c.contact_id
+		 AND attribute_type_id = @medicationstaking 
+		ORDER BY Start_Date DESC) AS medicationstaking,
+
+   (SELECT top 1 response
+	   FROM [ministryplatform].[dbo].[form_response_answers] r
+	   join (select fra.Form_Response_Answer_ID as max_id
```
chris-branch: Is the self-join needed?  It already has TOP 1 with ORDER BY Form_Response_Answer_ID, so doesn't seem like the self-join changes the outcome, but maybe I'm missing something?

PR #4298, opened 2017-08-08
US3747-Inactive-Login

abeeku23:
Removed the function I created to write changes to the audit log since the RESTApi automatically generates one on ".UpdateRecord()"


abeeku23:
This was completed with DE4069

PR #4327, opened 2017-08-17
us8782 try a group

plachmann:


Gateway/crds-angular/Services/GroupToolService.cs
```diff
@@ -347,27 +363,29 @@ public MyGroup GetMyGroupInfo(string token, int groupId)
             };
         }
 
-        public void ApproveDenyInquiryFromMyGroup(string token, int groupTypeId, int groupId, bool approve, Inquiry inquiry, string message = null)
+        public void ApproveDenyInquiryFromMyGroup(string token, int groupId, bool approve, Inquiry inquiry, string message, int roleId)
         {
             try
             {
-                var myGroup = GetMyGroupInfo(token, groupId);
+                // var myGroup = GetMyGroupInfo(token, groupId);
```
vchtelmakh: Does this need to be deleted?

plachmann: Yes. Deleted and pushed.

PR #4340, opened 2017-08-22
de4135 don't send old email on 'try a group'

plachmann:


Gateway/crds-angular/Services/GroupToolService.cs
```diff
@@ -836,20 +844,23 @@ public void SubmitInquiry(string token, int groupId)
             var props = new EventProperties {{"Name", group.GroupName}, {"City", group?.Address?.City}, {"State", group?.Address?.State}, {"Zip", group?.Address?.PostalCode}};
             _analyticsService.Track(contact.Contact_ID.ToString(), "RequestedToJoinGroup", props);
 
-            foreach (var leader in leaders)
+            if (sendEmail)
```
vchtelmakh: doSendEmail so it's obvious its bool

PR #4343, opened 2017-08-23
DE4151 only show deez docs when necessary

andycanterbury:


crossroads.net/app/trips/mytrips/mytripCard.html
```diff
@@ -7,7 +7,7 @@ <h3 class="card-row flush-bottom">{{trip.tripName}}</h3>
   </div><!-- /card-heading -->
 
   <!-- Documents -->
-  <div class="col-xs-12 documents" ng-show="hasDocuments(waivers) || showIPromise(trip.contactId)">
```
silbermm: it appears that this is the last use of `hasDocuments` - Should we remove it from the directive since it's not being used?

PR #4355, opened 2017-08-30
Feature/us9525 update embedded giving url

IngageKrizJ:
**DO NOT MERGE UNTIL SPRINT Z2!**

# Purpose
[US9525](https://rally1.rallydev.com/#/38108142417d/detail/userstory/144103845156): Updating the embed component (and components/controllers that use it) to support embedding things other than /give.

# Changes
This requires several minor updates in code for Camp Payments, Invoices, and the actual Embed component.  It also requires updates to several pages in CMS (the Live Stream page and the Online Giving page).  Those updates will be done manually, as we have not found a way to successfully script Page updates in the CMS.

# Other
This has a related pull request in [crds-embed](https://github.com/crdschurch/crds-embed/pull/308).


crossroads.net/core/embed/embed.controller.js
```diff
@@ -37,23 +56,39 @@ export default class EmbedController {
...
   }
 
+  $onDestroy() {
+    this.closeIframes();
+  }
+
   buildUrl() {
-    let path;
-    if (this.attrs.map === 'true') {
-      path = '/add-me-to-the-map?type=donation';
```
silbermm: is the `map` attribute not supported any more?

IngageKrizJ: No - that was a temporary for the "add me to the map" functionality, before connect launched.  I actually don't think it was ever used.

silbermm: :+1:  for deleting old code!

PR #4371, opened 2017-09-06
US7701 - update lat/long with profile update

plachmann:


Gateway/crds-angular.test/Services/PersonServiceTest.cs
```diff
@@ -1,7 +1,8 @@
 using System;
 using System.Collections.Generic;
-using System.Reactive.Linq;
+using System.Device.Location;
```
JoeKer1: I don't think this should be here.

plachmann: tried to remove it. It was pulled in by the GeoCoordinate object

JoeKer1: ahh my bad then 👍

PR #4376, opened 2017-09-07
DE4233 group searchable after delete

vchtelmakh:


Gateway/crds-angular/Services/GroupToolService.cs
```diff
@@ -678,6 +681,8 @@ public bool ValidateUserAsLeader(string token, int groupTypeId, int groupId, int
 
         public void EndGroup(int groupId, int reasonEndedId)
         {
+            _awsCloudsearchService.DeleteGroupFromAws(groupId);
```
JoeKer1: Should this happen after the group is ended in MP?

vchtelmakh: It should be async and happen simultaneously. Want me to do that? 

Refreshes every 5 mins anyway.

JoeKer1: Makes sense. I don't see anything marked as async though. Probably not a big deal since the job will run and it'll be back in AWS after that.

vchtelmakh: This is a moot point now since it's merged, but I mean that, theoretically, we should validate the data and then save to both places simultaneously to save time. Again, purely from a "best practice" perspective - async done right kind of takes over the app, so it would be a pain.

PR #4393, opened 2017-09-12
Feature/us9750 add and edit group immediately in aws

vchtelmakh:
---


JoeKer1:
Do we need to do all the time processing in AwsTimeHelper.cs? Can we just do that when we map from AWS ?

JoeKer1:
Nevermind that's where we use it

vchtelmakh:
We need it in a certain format in AWS. The problem is that we store a Time in SQL, but need a DateTime in AWS - so we are forced to use a placeholder date. AWS can only take a UTC string for DateTime. We do map it:

```
            Mapper.CreateMap<MpConnectAws, AwsConnectDto>()
                .ForMember(dest => dest.LatLong,
                           opts => opts.MapFrom(
                               src => (src.Latitude == null || src.Longitude == null) ? "0 , 0" : $"{src.Latitude} , {src.Longitude}"))
                .ForMember(dest => dest.GroupAgeRange,
                           opts => opts.MapFrom(
                               src => src.GroupAgeRange.Split(',')))
                .ForMember(dest => dest.GroupCategory,
                           opts => opts.MapFrom(
                               src => src.GroupCategory.Split(',')))
                .ForMember(dest => dest.GroupMeetingTime,
                           opts => opts.MapFrom(
                               src => new AwsTimeHelper().ConvertTimeToUtcString(src.GroupMeetingTime)));```

Didn't want to put too much logic in AutoMapper inline.

File: Gateway/crds-angular.test/Services/AwsCloudsearchServiceTest.cs [hunk 1]
```diff
@@ -24,7 +24,7 @@ public void SetUp()
             _mpFinderRepository = new Mock<IFinderRepository>();
             _mpConfigurationWrapper = new Mock<IConfigurationWrapper>();
 
-            _fixture = new AwsCloudsearchService(_addressGeocodingService.Object, _mpFinderRepository.Object, _mpConfigurationWrapper.Object);
+            _fixture = new AwsCloudsearchService(_mpFinderRepository.Object, _mpConfigurationWrapper.Object);
 
             //force AutoMapper to register
             AutoMapperConfig.RegisterMappings();
```
JoeKer1: Are the appropriate tests already here?

PR #4396, opened 2017-09-12
US9686: Make login faster

k8tied1:


Gateway/MinistryPlatform.Translation/Repositories/UserRepository.cs
```diff
@@ -103,11 +111,26 @@ public List<MpRoleDto> GetUserRoles(int userId)
...
+        }
+
         public void UpdateUser(Dictionary<string, object> userUpdateValues)
         {
             MinistryPlatformService.UpdateRecord(Convert.ToInt32(ConfigurationManager.AppSettings["Users"]), userUpdateValues, ApiLogin());
         }
 
+        public void UpdateUserRest(Dictionary<string, object> userUpdateValues, int user_id)
+        {
+            _ministryPlatformRest.UsingAuthenticationToken(ApiLogin()).UpdateRecord("dp_Users", user_id, userUpdateValues);
```
chris-branch: Beware: the 2nd parameter in UpdateRecord() is not actually used -- you must include the PK ID within the dictionary.  It might be a good idea to pass -1 for the 2nd param to make that clear to future readers.

Gateway/crds-angular/Services/LoginService.cs
```diff
@@ -114,6 +114,16 @@ public bool ClearResetToken(string username)
             userUpdateValues["User_ID"] = user_ID;
             userUpdateValues["ResetToken"] = null; // swap out for real implementation
             _userService.UpdateUser(userUpdateValues);
+            
+            return true;
+        }
+
+        public bool ClearResetToken(int userId) //KD use if we already have the id to save time
```
chris-branch: Could the ClearResetToken(string userName) overload call the new ClearResetToken(int userId) overload to eliminate some duplicate code?

Gateway/crds-angular/Controllers/API/LoginController.cs
```diff
@@ -137,36 +137,35 @@ public IHttpActionResult Post([FromBody] Credentials cred)
...
                 {
                     return this.Unauthorized();
                 }
-
-                var userRoles = _personService.GetLoggedInUserRoles(token);
-                var user = _userService.GetByAuthenticationToken(token);
-                var p = _personService.GetLoggedInUserProfile(token);
+                var user = _userService.GetByUserName(cred.username); //235 ms _userService.GetByAuthenticationToken(token) was 1.5 seconds 
+                var userRoles = _userService.GetUserRolesRest(user.UserRecordId);
+                var c = _contactRepository.GetContactByUserRecordId(user.UserRecordId);//use a rest call and use the id directly
```
chris-branch: GetContactByUserRecordId can fail if there are multiple records (i.e., multiple Contacts that share the same User_Account).  Looks like there are 5 instances of that in prod right now.  Not sure if this is a concern, but just FYI it's a change in behavior.

Gateway/MinistryPlatform.Translation/Repositories/GetMyRecordsRepository..cs
```diff
@@ -5,6 +5,7 @@
 using Crossroads.Utilities.Interfaces;
 using Crossroads.Web.Common;
 using Crossroads.Web.Common.Configuration;
+using Crossroads.Web.Common.MinistryPlatform;
```
chris-branch: Looks like this file could be excluded from the commit ?

Gateway/crds-angular/Services/LoginService.cs
```diff
@@ -118,6 +118,16 @@ public bool ClearResetToken(string username)
             return true;
         }
 
+        public bool ClearResetToken(int userId) //KD use if we already have the id to save time
+        {
+            Dictionary<string, object> userUpdateValues = new Dictionary<string, object>();
+            userUpdateValues["User_ID"] = userId;
+            userUpdateValues["ResetToken"] = null; // swap out for real implementation
```
Ganonside: This would look cleaner if the Dictionary object initializer was used

Gateway/crds-angular/Controllers/API/LoginController.cs
k8tied1: GetMyProfile throws an exception if there are multiple records as well. *Checking current behavior*

k8tied1: The behavior would be the same - I tested this in int and got the error 

message: "Login Failed", errors: ["GetMyProfile returned multiple records"]

Gateway/MinistryPlatform.Translation/Repositories/UserRepository.cs
```diff
@@ -62,6 +63,15 @@ public MpUser GetByAuthenticationToken(string authToken)
             return user;
         }
 
+        public MpUser GetByUserName(string userName)
+        {
+            string userNameClean = Regex.Replace(userName, @"[\r\n\x00\x1a\\'""]", @"\$0"); //prevent SQL injection
```
IngageKrizJ: This would be nice to get into somewhere standard - not saying it needs to be done on this change, but we have this same potential issue all over the place (as Chris has pointed out).

Gateway/crds-angular/Controllers/API/LoginController.cs
```diff
@@ -137,36 +137,35 @@ public IHttpActionResult Post([FromBody] Credentials cred)
...
+                
                 if (token == "")
                 {
                     return this.Unauthorized();
                 }
-
-                var userRoles = _personService.GetLoggedInUserRoles(token);
-                var user = _userService.GetByAuthenticationToken(token);
-                var p = _personService.GetLoggedInUserProfile(token);
+                var user = _userService.GetByUserName(cred.username); //235 ms _userService.GetByAuthenticationToken(token) was 1.5 seconds
```
IngageKrizJ: All three of these calls do an ApiLogin() - if further improvements are needed, could we do the ApiLogin() once here, and pass the token in on each call?  I don't love the idea of doing that, but you could make the token an optional parameter in each of those calls, and do ApiLogin() only if it is null, otherwise use the given token.

Gateway/MinistryPlatform.Translation/Repositories/UserRepository.cs
```diff
@@ -64,8 +65,9 @@ public MpUser GetByAuthenticationToken(string authToken)
 
         public MpUser GetByUserName(string userName)
         {
-            string searchUser = $"dp_Users.User_Name='{userName}'";
-            string columns = "User_Name,User_GUID, Can_Impersonate,User_Email,User_ID";
+            string userNameClean = Regex.Replace(userName, @"[\r\n\x00\x1a\\'""]", @"\$0"); //prevent SQL injection
```
chris-branch: To escape quotes for SQL, we would need to double up the quotes instead of adding a backslash.  Something like: userName.Replace("'", "''")

PR #4405, opened 2017-09-14
Add new line to set Available_Online flag

vchtelmakh:
Added line 67 to existing stored proc - it was missing field: 
`(IIF(G.Available_Online IS NULL, 0, G.Available_Online)) AS groupavailableonline`


JoeKer1:
Available online does not mean it's an online group, only that it will show in search. Online groups have this set to true and also do not have an address.

vchtelmakh:
Changed the title

PR #4406, opened 2017-09-14
consolidated calls into analytics service

Doug-Shannon:
Moved all analytics calls in crds-angular into analytics service and added test coverage.


Doug-Shannon:
Dont merge yet

Doug-Shannon:
It can be merged now

PR #4411, opened 2017-09-15
Defect/de4291 capture contact

ilpadre:


Gateway/crds-angular/Controllers/API/UserController.cs
```diff
@@ -46,7 +47,9 @@ public IHttpActionResult Post([FromBody] User user)
             try
             {
                 var userRecord = _accountService.RegisterPerson(user);
-                _analyticsService.Track(userRecord.email, "SignedUp");
+                var user_ID = _userRepository.GetUserIdByUsername(user.email);
+                var contact_Id = _userRepository.GetContactIdByUserId(user_ID);
```
Doug-Shannon: I know this is the only way to get it, but now we are making 2 extra calls per signin to get that data.  Are we ok with that?

PR #4420, opened 2017-09-20
Two files to add test data for Mp To Gp testing

candrews-thrive:


CI/TestSQL/02.TestData/CreateDepositForDonationAndPayment.sql
```diff
@@ -0,0 +1,103 @@
+USE [MinistryPlatform]
+GO
+
+-- Get donor records
+DECLARE @benDonorID as int
+SET @benDonorID = (SELECT Donor_ID FROM Donors WHERE Contact_ID = (SELECT Contact_ID FROM Contacts WHERE First_Name = "Ben" AND Last_Name = "Kenobi"))
```
s-henney: This won't change Ben's status as a previous CC donor, will it?

candrews-thrive: I don't think so. The payment type for his donation is 4 which is CC.

candrews-thrive:
Hold off on merging this I have to make some changes to the scripts

PR #4423, opened 2017-09-21
US9852 - add user different leader email

plachmann:


Gateway/crds-angular/Controllers/API/FinderController.cs
```diff
@@ -30,7 +29,7 @@ public class FinderController : MPAuth
         private readonly IAuthenticationRepository _authenticationRepo;
         private readonly IAnalyticsService _analyticsService;
         private readonly ILog _logger = LogManager.GetLogger(MethodBase.GetCurrentMethod().DeclaringType);
-        private const int _trialMemberRoleId = 67;
+        private const int TrialMemberRoleId = 67;
```
JoeKer1: Should that number go in the MinistryPlatform config thing?

JoeKer1: Actually I think it's already there should we get it from there in the 0% chance we change the id?

Gateway/crds-angular/Services/FinderService.cs
```diff
@@ -1223,53 +1282,6 @@ public void SendEmailToAddedUser(string token, User user, int groupid)
             RecordCommunication(connection);
         }
        
-        private List<PinDto> TransformGroupDtoToPinDto(List<GroupDTO> groupDTOs, string finderType)
```
JoeKer1: did we not use this?

plachmann: VS showed it as unused so I removed it.

PR #4433, opened 2017-09-25
Fix to test teardown script

s-henney:
Script now removes any events manually added to the Go Midgar program.

### DIFF
--- a/CI/TestSQL_TearDown/02.Teardown_MPData/MidgarTrip.sql
+++ b/CI/TestSQL_TearDown/02.Teardown_MPData/MidgarTrip.sql
@@ -75,9 +75,13 @@ delete from event_groups where event_id in (select event_id from events where pr
 
 delete from event_equipment where event_id in (select event_id from events where program_id in (select program_id from programs where program_name = @tripName));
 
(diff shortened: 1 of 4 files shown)


CI/TestSQL_TearDown/02.Teardown_MPData/MidgarTrip.sql
```diff
@@ -2,7 +2,7 @@ USE [MinistryPlatform]
 GO
 
 --Retrieve name of trip in case create occurred in a year prior to teardown
-DECLARE @tripName AS VARCHAR(18)
+DECLARE @tripName AS VARCHAR(24)
```
candrews-thrive: Does it make sense to increase varchar to something like 50? I am thinking putting it that high will give us more room in case the results below return a program name is bigger the 24.

lippsm: If you decide to go this route (increase to 50), consider changing the Go Bedrock trip as well. Also, this needs to be changed in the setup script as well as the teardown.

s-henney: Bah, humbug. Gimme a minute.

PR #4466, opened 2017-10-06
us9940 - AddParticipantsToGroup and CreateGroupInquiry now make REST calls

plachmann:


Gateway/MinistryPlatform.Translation/Repositories/GroupRepository.cs
```diff
@@ -167,43 +163,42 @@ public bool GetDoesUserLeadSomeGroup(int userParticipantId)
             return doesUserLeadSomeGroup; 
         }
 
-        public int addParticipantToGroup(int participantId,
+        public int AddParticipantToGroup(int participantId,
                                          int groupId,
                                          int groupRoleId,
-                                         Boolean childCareNeeded,
+                                         bool childCareNeeded,
+                                         bool employeeRole,
```
JoeKer1: did this not have to be defaulted to false?

Gateway/MinistryPlatform.Translation/Repositories/GroupRepository.cs
```diff
@@ -1029,32 +1024,15 @@ public int UpdateGroup(MpGroup group)
...
-                {"Group_ID", inquiry.GroupId},
-                {"Email", inquiry.EmailAddress},
-                {"Phone", inquiry.PhoneNumber},
-                {"First_Name", inquiry.FirstName},
-                {"Last_Name", inquiry.LastName},
-                {"Inquiry_Date", System.DateTime.Now}
-            };
-
-            WithApiLogin<int>(token =>
+                _ministryPlatformRestRepository.UsingAuthenticationToken(token).Post<MpInquiry>(new List<MpInquiry> {inquiry});
```
JoeKer1: Love how much simpler this is :D

Gateway/MinistryPlatform.Translation/Repositories/GroupRepository.cs
plachmann: most call were providing a value, so I eliminated the default to better order and cleanup the method signature.

PR #4481, opened 2017-10-13
Feature/us10270-Henney Add user to map script

s-henney:
Script to add users to map given file of email-password combos and file of email-address combos.
Functionality should be the same as createMapUsers except this doesn't create a new user account.


s-henney:
Branch naming is hard.

PR #4482, opened 2017-10-13
Feature/us10270-Henney Add user to map script 

s-henney:
Script to add users to map given file of email-password combos and file of email-address combos.
Functionality should be the same as createMapUsers except this doesn't create a new user account.

### DIFF
--- a/CI/TestSQL/01.TestUsers/addUsersToMap.ps1
+++ b/CI/TestSQL/01.TestUsers/addUsersToMap.ps1
@@ -8,10 +8,9 @@ param (
     [string]$uploadAwsEndpoint = 'https://gatewayint.crossroads.net/gateway/api/finder/uploadallcloudsearchrecords'
  )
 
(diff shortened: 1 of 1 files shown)


s-henney:
Branch naming is hard.

CI/TestSQL/01.TestUsers/addUsersToMap.ps1
```diff
@@ -0,0 +1,121 @@
...
+				updateHomeAddress = "false"
+				firstName = $mapUser.first
+				lastName = $mapUser.last
+				#siteName = [NullString]::Value
+				emailAddress = $mapUser.email
+				contactId = $participantRecord.ContactId
+				participantId = $participantRecord.ParticipantId
+				address = @{
+					#addressId = [NullString]::Value
+					addressLine1 = $mapUser.address1
```
simeonsena: address has to be a real address recognized by google maps in order for the long/lat to be found. Are we always using real addresses?

PR #4483, opened 2017-10-13
updated START_DATE AS STARTDATE to just START_DATE - also threw error…

Doug-Shannon:
… back to front end.


Gateway/crds-angular/Services/GroupService.cs
```diff
@@ -913,6 +913,7 @@ public void UpdateGroupParticipantRole(int groupId, int participantId, int roleI
             {
                 var message = $"Could not update group participant {participantId}";
                 _logger.Error(message, e);
+                throw e;
```
JoeKer1: good job throwing this

PR #4504, opened 2017-10-26
US10378 - Reskin /serve

seancdavis:
Remove the fluid container, which added unnecessary padding.

I'm worried this change will introduce regressions on other pages using the Screen Width template, but a list of which pages those are isn't readily available, so I'm not exactly sure how we go about regression testing.

Also note that, for now, the testing should be done at `/ddk-reskin/serve`. The main serve page will be changed after this is merged and after the proper channels are notified.

### DIFF
--- a/crossroads.net/core/templates/screenWidth.html
+++ b/crossroads.net/core/templates/screenWidth.html
@@ -1,18 +1,13 @@
 <ng-include src="'templates/header.html'"></ng-include>
 
 <div ng-class="{'crds-legacy-styles': renderLegacyStyles, 'crds-styles': !renderLegacyStyles}">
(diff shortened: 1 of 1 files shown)


zarahzachz:
@seancdavis It looks like there's some padding problems on mobile screens. 

screen shot 2017-10-27 at 11 55 46 am

seancdavis:
@zarahzachz: Are you viewing through Maestro and did you rebuild crds-angular? (I'm not seeing this issue.)

screen shot 2017-10-30 at 4 53 26 pm

zarahzachz:
@seancdavis Well shoot - you're right, I didn't check it in maestro. Merging!

PR #4569, opened 2017-11-30
US10663 - Signin Validation

seancdavis:
Touch all fields in /signin and /register when submitting without entering data.


zarahzachz:
@seancdavis It looks like some of the tests are failing.

Otherwise, signin looks great. I did notice that the registration page is still behaving incorrectly. I checked the story out and it looks like I failed to add proper AC covering that part of it. Once the tests are fixed, I'm okay with merging and creating a new story to govern the registration page unless you'd be okay with adding the needed registration page changes to this story. Let me know which you'd prefer.

seancdavis:
@zarahzachz: Nah, I had register fixed and then made a change without retesting and it broke it again. Let's see if this fixes the specs.

PR #4578, opened 2017-12-07
US10939 journey view

plachmann:


CI/SQL/20171207_090000_US10939-onsiteandoffsitegroups-function.sql
```diff
@@ -0,0 +1,39 @@
...
+		FROM group_participants GP
+		JOIN GROUPS G ON G.Group_ID = GP.Group_ID
+		WHERE (GP.end_date IS NULL OR GP.end_date > GETDATE()) AND GP.participant_id = @ParticipantId
+		AND G.Group_Type_ID = 1); -- small group
+
+    SET @OnsiteGroupCount = (select COUNT(*) 
+		FROM group_participants GP
+		JOIN GROUPS G ON G.Group_ID = GP.Group_ID
+		WHERE (GP.end_date IS NULL OR GP.end_date > GETDATE()) AND GP.participant_id = @ParticipantId
+		AND G.Group_Type_ID = 8); -- onsite group
```
JoeKer1: I don't know that they'll actually use the onsite group type to house these groups unfortunately. I think they may want them to show up on the group tool. Not 100% sure how this is going to work yet. Based on a short conversation with Kelley

JoeKer1:
I think we should also remove all the existing journey views, but maybe we do that manually.

PR #4599, opened 2018-01-02
Feature/us10977 suppress reg email for xroads app

Ganonside:


Gateway/crds-angular/Controllers/API/UserController.cs
```diff
@@ -42,11 +42,11 @@ public class UserController : MPAuth
         [VersionedRoute(template: "user", minimumVersion: "1.0.0")]
         [Route("user")]
         [HttpPost]
-        public IHttpActionResult Post([FromBody] User user)
+        public IHttpActionResult Post([FromBody] User user, int? householdSourceId)
```
chris-branch: Add a default value ( = null) for householdSourceId so that the route works when the parameter is not specified.

crossroads.net/core/services/user_service.js
```diff
@@ -3,10 +3,10 @@
         factory('User', ['$resource','$log', UserService]);
 
     //By declaring api/user a resource, angular provides us with helpful verbs to perform CRUD operations. (save/update)
- 
+
 
     function UserService($resource, $log) {
-        var User = $resource(__GATEWAY_CLIENT_ENDPOINT__ + 'api/user');
+        var User = $resource(__GATEWAY_CLIENT_ENDPOINT__ + 'api/user?householdSourceId=');
```
chris-branch: Revert this (no change needed here once you have a default value so the parameter becomes optional)

CI/SQL/20180102_113000_US10977-add-mobile-app-household-source.sql
```diff
@@ -0,0 +1,59 @@
...
+-- Date: 1/2/2018
+-- Description: Add a Household Source for the mobile
+-- 	app and stop sending the default welcome email to
+--	to users who sign up from it.
+-- =================================
+
+DECLARE @mobileAppHouseholdSourceId INT = 51; -- Not in MPIdentityManagement Range
+DECLARE @domainId INT = 1;
+
+BEGIN
```
chris-branch: BEGIN / END scope is wrong.  Should be IF .... BEGIN .... END

CI/SQL/20180102_113000_US10977-add-mobile-app-household-source.sql
```diff
@@ -0,0 +1,59 @@
...
+	)
+
+	SET IDENTITY_INSERT Household_Sources OFF;
+END
+
+DECLARE @defaultWelcomeEmailProcessId INT = 23;
+DECLARE @previousDependentCondition nvarchar(4000);
+DECLARE @newDependentCondition nvarchar(4000);
+
+BEGIN
```
chris-branch: BEGIN / END scope is wrong.  Should be IF .... BEGIN .... END

CI/SQL/20180102_113000_US10977-add-mobile-app-household-source.sql
Ganonside: done

CI/SQL/20180102_113000_US10977-add-mobile-app-household-source.sql
Ganonside: done

PR #4622, opened 2018-01-17
Feature/us11833 Fixed reload issues

s-henney:
-Updated test data reload scripts to find contact id from user account (user names are guaranteed unique, while contact's emails are not). Should prevent "query returned more than one result" related issues in reload scripts.
-Saved group ids instead of re-finding them for each group participant addition
-Added UpdateUserData.ps1 and some CSVs as setup for better reload scripts

TC Build:
http://mp-ci.centralus.cloudapp.azure.com/viewLog.html?buildId=40559&buildTypeId=Qa_Release_TeardownAndReloadTestData&tab=buildLog#_focus=191&state=191

TODO:
-There are 2 *Family.sql scripts and SignUpToServe_GroupParticipants.sql that are failing. Both are caused because contacts are not being torn down and will be fixed in a future PR.


CI/TestSQL/01.TestUsers/UpdateContact.csv
```diff
@@ -0,0 +1,4 @@
+﻿User_Email,Middle_Name,Date_of_Birth,Gender_ID,Marital_Status_ID,Household_Position_ID,Mobile_Phone,Company_Phone,Prefix_ID,,
```
candrews-thrive: Is this CSV file and the ones below to be used instead of rolling our own SQL scripts?

s-henney: @candrews-thrive That's the idea. We're duplicating way more code than we need to by rolling our own queries, and most of it is copy-pasta. Building common data through CSVs should make it easier to maintain queries and for anyone to add new test data.
I didn't mean to include these changes in this PR, so it'll be improved in a future PR.

candrews-thrive: Awesome

PR #4638, opened 2018-01-31
Feature/us10014 better form submission errors

IngageRitterS:


CI/SQL/20180126_132605_US10014-BetterFormSubmissionErrors.sql
```diff
@@ -0,0 +1,30 @@
+USE MinistryPlatform
+GO
+
+DECLARE @Template_ID int = 2035;
+DECLARE @From_Contact_ID int = 1;
```
vchtelmakh: Should we somehow indicate that this is the "Administrator, MinistryPlatform" contact? Don't need to change, just a thought.

PR #4660, opened 2018-02-09
if invoice exists and camp registration not finalized, use amt from …

IngageRitterS:
if invoice exists and camp registration not finalized, use amt from invoice not product base price


Gateway/crds-angular.test/Services/CampServiceTest.cs
```diff
@@ -1139,9 +1139,71 @@ public void ShouldGetProductInfo()
...
+        public void ShouldGetProductInfoAndUseInvoiceTotal()
+        {
+            const int eventId = 1234;
+            const string token = "1aaaa";
+            var product = new MpProduct
+            {
+                ProductId = 111,
+                BasePrice = 2000,
+                DepositPrice = 200,
+                ProductName = "Hipster Beard Wax"
```
andycanterbury: 🤣

PR #4664, opened 2018-02-10
feature/US14231

simeonsena:
Added participant to connect group for automation


simeonsena:
per @s-henney, this user has been added to the CSV file. This PR is now obsolete

PR #4672, opened 2018-02-14
if there are multiple contact records look at users

JoeKer1:


Gateway/MinistryPlatform.Translation/Repositories/ContactRepository.cs
```diff
@@ -253,7 +253,15 @@ public int GetActiveContactIdByEmail(string email)
...
             if (records.Count > 1)
             {
-                throw new Exception("User email did not return exactly one user record");
+                _logger.Error($"There are multiple contacts for email: <{email}>, trying to find a unique contact by user record");
+                var userRecord =
+                    _ministryPlatformRest.UsingAuthenticationToken(token)
+                        .Search<MpUser, MpContact>($"User_Name = '{email}'", "Contact_Id");
+                if (userRecord.Count < 1)
+                {
+                    throw new Exception("User email did not return exactly one contact record and no user record found");
```
JoeKer1: I think I should throw here so we don't add to the multiple contacts problem ?

PR #4691, opened 2018-03-07
feature/us12539 - TD/RL better error handling

s-henney:
- Previously, if an entry could not be created in MP, the remaining entries in the same file would be skipped. With this update, the script will attempt to create all data in the file and report a failing exit code at the end if anything could not be created.
- The master file called by TC will now report a failing exit code if any entry could not be created so TC can be configured to stop the build.

[TeamCity run ](http://mp-ci.centralus.cloudapp.azure.com/viewLog.html?buildId=44740&buildTypeId=Qa_Release_TeardownAndReloadTestData&tab=buildLog#_state=64)where failed teardown stops further steps.

Bonus:
- Fixed a group attribute configuration

Todo:
- The failure is an actual bug that needs to be fixed, but it does demo the exit codes added in this PR work. Bug will be fixed in a future PR.

### DIFF
--- a/CI/TestSQL/00.PowershellScripts/DeleteTestData.ps1
+++ b/CI/TestSQL/00.PowershellScripts/DeleteTestData.ps1
@@ -7,14 +7,25 @@ param (
 #Call scripts in subfolders relative to current script
 $root_path = (Split-Path (Split-Path ($MyInvocation.MyCommand.Definition)))
 
-
+$errors = 0
 #Delete from UserData folder
 . "$($root_path)\01.TestUsers\DeleteTestUsers.ps1" -DBServer $DBServer -DBUser $DBUser -DBPassword $DBPassword
+$errors += $LASTERRORCODE
 
 #Delete from TestData folder
 . "$($root_path)\02.TestData\01.Batch+Deposit\DeleteDepositsAndBatches.ps1" -DBServer $DBServer -DBUser $DBUser -DBPassword $DBPassword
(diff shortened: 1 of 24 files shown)


CI/TestSQL/00.PowershellScripts/DeleteTestData.ps1
```diff
@@ -7,14 +7,25 @@ param (
 #Call scripts in subfolders relative to current script
 $root_path = (Split-Path (Split-Path ($MyInvocation.MyCommand.Definition)))
 
-
+$errors = 0
 #Delete from UserData folder
 . "$($root_path)\01.TestUsers\DeleteTestUsers.ps1" -DBServer $DBServer -DBUser $DBUser -DBPassword $DBPassword
+$errors += $LASTEXITCODE
```
candrews-thrive: Is exit code a number or a string? Would it be useful to have each failure's code to debug what failed?

s-henney: It stores a number that powershell automatically sets it when an exit code is returned.
I don't think passing unique error codes to this script or the CreateTestData script would be very valuable since error details are already reported when encountered in the lower script levels. If a create/delete script fails due to data issues, a message from the SQL script is reported with enough data to start debugging, and if powershell can't execute a command or connect to the database it's also caught and the stacktrace is reported (that's the "$_" in the catch block).
And we'd want TC to stop the build in any of the error scenarios, not just some.

PR #4705, opened 2018-03-16
scroll to topmost error in profile

JoeKer1:
Desktop: 
profileerror

Mobile: 
profilemobile

*Note: Toast looked like that before this PR.


JoeKer1:
Only has the scroll to topmost error logic

PR #4718, opened 2018-04-04
DE5238 add code to check start date

ilpadre:


Gateway/crds-angular/Services/EmailCommunication.cs
```diff
@@ -114,14 +114,22 @@ public void SendEmail(EmailCommunicationDTO email, string token)
                 communication.MergeData.Add("BaseUrl", _configurationWrapper.GetConfigValue("BaseUrl"));
             }
 
-            var communicationId = _communicationService.SendMessage(communication, true);
-            var directEmailId = new DirectEmailCommunication()
+            bool sendImmediately = communication.StartDate <= DateTime.Today;
```
chris-branch: I don't think this will work.  `DateTime.Today` has the time set to 00:00:00, so `DateTime.Now <= DateTime.Today` will rarely be true.  I think we'd be better with something like `bool sendImmediately = email.StartDate == null;`

PR #4723, opened 2018-04-06
US12746: use $window.location for absolute urls for redirecting via cookie

timmyg:
The redirectUrl cookie currently does not support going to absolute URL's, so checking to see if string starts with http/https and using $window.location to go directly to it.

### DIFF
--- a/crossroads.net/core/services/session_service.js
+++ b/crossroads.net/core/services/session_service.js
@@ -37,6 +37,7 @@
     $http,
     $state,
     $location,
+    $window,
     $interval,
     $timeout,
     $cookies,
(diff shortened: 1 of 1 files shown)


crossroads.net/core/services/session_service.js
```diff
@@ -157,17 +158,24 @@
           const params = vm.exists('params');
           vm.removeRedirectRoute();
 
-          const foundState = $state.get().filter(state => state.name === url);
-          if (foundState.length > 0) {
-            if (params === undefined) {
-              $state.go(url);
+          // check if absolute url
+          var absoluteUrlRegex = /^https?:\/\//i;
```
vchtelmakh: All major browsers support `let` now.

timmyg: good call

PR #4745, opened 2018-04-19
update hotfix file

tatelucas:


crossroads.net/styles/hotfix-seventy-five.css
```diff
@@ -64,15 +64,13 @@
...
     position: absolute;
     left: 50%;
     top: 1%;
     transform: translate(-50%,1%);
     width: 100%;
   }
-
 }
 
 @media (min-width:768px) and (max-width:992) {
```
lukemoderwell: missing a `px` here (992)

tatelucas: Good catch! Just fixed it.

PR #4759, opened 2018-04-30
MP Winter Upgrade changes

timmyg:
This PR includes changes to the crds-angular project to support the MP Winter 2018 upgrade that went in on April 30, 2018. crds-angular now consumes Crossroads.Web.Common and Crossroads.ClientApiKeys nuget packages v2.0.0.


Gateway/MinistryPlatform.Translation.Test/Services/MinistryPlatformRestServiceIntegrationTest.cs
```diff
@@ -26,10 +27,17 @@ public class MinistryPlatformRestServiceIntegrationTest
         public void SetupAll()
         {
             var config = new ConfigurationWrapper();
-            var authenticationRepository = new AuthenticationRepository(new RestClient(config.GetEnvironmentVarAsString("MP_OAUTH_BASE_URL")),
-                                                                        new RestClient(config.GetEnvironmentVarAsString("MP_REST_API_ENDPOINT")));
-            var ministryPlatformRestClient = new RestClient(config.GetEnvironmentVarAsString("MP_REST_API_ENDPOINT"));
-            var apiUserRepository = new ApiUserRepository(config, authenticationRepository, ministryPlatformRestClient);
+            var authRestClient = new HttpClient
+            {
+                BaseAddress = new Uri(config.GetEnvironmentVarAsString("MP_OAUTH_BASE_URL"))
```
timmyg: were these values update in google drive document?

timmyg:
Looks good to me.

- /MinistryPlatform.Models/packages.config should we update Newtownsoft.Json to match the other version upgrades (10.0.3)?
- I don't fully understand why the projects are now including System.Security.Cryptography.* as I dont see any changes in the code around it, but I am sure I am missing something
- The Crossroads.Constants change is a bit confusing as I dont understand how it was working before, but again, I'm sure im missing something

mjvcallibrity:
I can't speak to the Cryptography changes (best guess is it has something to do with one of the other dependent packages that we had to update, or maybe because we are now pointing to framework v4.6.1), but the Crossroads.Constants change deals with the fact that when we updated the framework version as well as web.common and clientapikeys, we were getting an ambiguous reference to `Constants` in several of our files and we needed to explicitly point to which Constants we wanted referenced.

mjvcallibrity:
@timmyg I'm not seeing a `/MinistryPlatform.Models/packages.config` file anywhere. It feels like I'm missing something obvious.

mjvcallibrity:
Given the above explanations, @timmyg @chris-branch , are the both of you good to merge this, or is there anyone else we want to pull into this?

mjvcallibrity:
I'm going to go ahead and merge. There is much developer grumbling right now.

PR #4773, opened 2018-05-04
add scripted NOLA trip and add id to i promise button

ajcolub11:
"Scripted" a NOLA Trip
added an ID to the my trips page


crossroads.net/app/trips/mytrips/mytripCard.html
```diff
@@ -12,7 +12,7 @@ <h5 class="flush-bottom">Sign These Documents</h5>
 
     <!-- I Promise -->
     <div class="go-trip-document" ng-show="showIPromise(trip.contactId)">
-      <button class="btn btn-default go-trip-document-btn" ng-disabled="trip.iPromiseSigned" ng-click="signIPromise(trip.eventParticipantId)">
+      <button class="btn btn-default go-trip-document-btn" ng-disabled="trip.iPromiseSigned" ng-click="signIPromise(trip.eventParticipantId)" id="signIPromiseButton">
```
s-henney: Could you use data-automation-id instead of id here?

PR #4797, opened 2018-05-23
Defect/de5476 handle no program

IngageRitterS:


Gateway/crds-angular.test/Services/PaymentServiceTest.cs
```diff
@@ -772,6 +773,11 @@ private static MpProgram fakeProgram2(int productId)
           };
         }
 
+        private static MpProgram fakeProgram3(int productId)
+        {
+          return new MpProgram();
```
andycanterbury: I get that you were following the established pattern, but I feel this could have just been done inline.

PR #4861, opened 2018-07-17
Added two scripts to create and delete testing printer map

candrews-thrive:


CI/TestSQL/CustomScripts/Reload/CreatePrinterMap.sql
```diff
@@ -0,0 +1,5 @@
+USE [MinistryPlatform]
+GO
+
+INSERT INTO cr_Printer_Maps (Printer_ID,Printer_Name       ,Computer_ID,Computer_Name      ,Domain_ID)
```
s-henney: Can you check that this printer map does not exist already before you add it?

CI/TestSQL/CustomScripts/Teardown/DeletePrinterMap.sql
```diff
@@ -0,0 +1,12 @@
...
+GO
+
+DECLARE @printerMapID as int
+SET @printerMapID = (SELECT Printer_Map_ID FROM cr_Printer_Maps WHERE Printer_ID = 0
+								AND Printer_Name = '(t+auto) Printer'
+								AND Computer_ID = 0
+								AND Computer_Name = '(t+auto) Computer'
+		    )
+
+DELETE FROM cr_Kiosk_Configs WHERE Printer_Map_ID = @printerMapID
```
s-henney: Can you check to make sure @printerMapID is not null before deleting these? Pretty sure SQL will throw an error if it sees "= null"

candrews-thrive:
Address the PR comments

PR #4882, opened 2018-07-31
Defect/de5698 fix hi

Doug-Shannon:


Doug-Shannon:
in the finder service layer I guess my machine linted the whole thing - so it looks like a lot was changed.

The only changes are in the SayHi and the SendSayHi functions.  I also added code at the top to get the correct email templates.

PR #4925, opened 2018-08-31
Feature/US14700 airport selection on my trips dashboard

vchtelmakh:
---


crossroads.net/app/trips/preferred-airport.service.js
```diff
@@ -0,0 +1,93 @@
+(function() {
+  'use strict';
+  module.exports = PreferredAirportService;
+
+  function PreferredAirportService() {
+    const consts = require('crds-constants');
+    const preferredAirportAttributeTypeId = consts.ATTRIBUTE_TYPE_IDS.PREFERRED_AIRPORT;
+
+    const COMMON_AIRPORT_NAMES = {
```
IngageRitterS: Why aren't these constants stored in constants.js?

vchtelmakh: Well, it's because [bs reason here...], changed =)

PR #4944, opened 2018-09-19
remove ability to manage recurring gifts, add message about pushpay blackout

timmyg:
do not delete branch after merge as this may be a hotfix on 9/26

### DIFF
--- a/crossroads.net/app/profile/giving/profileGiving.html
+++ b/crossroads.net/app/profile/giving/profileGiving.html
@@ -43,6 +43,9 @@
             <div class="row recurring-gifts push-bottom">
                 <div class="col-sm-12">
                     <h2>Recurring Gifts</h2>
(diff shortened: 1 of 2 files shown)


crossroads.net/app/profile/giving/profileGiving.html
```diff
@@ -43,6 +43,10 @@ <h2>Commitments</h2>
             <div class="row recurring-gifts push-bottom">
                 <div class="col-sm-12">
                     <h2>Recurring Gifts</h2>
+                    <div class="alert alert-info" role="alert">
```
jpc11623: Would there be any value in making this into a content block?

briancavanaugh: Possibly....but...this will be very short lived.  So not sure it's worth it?

PR #4945, opened 2018-09-20
US14663 add sp to set HH Pos

ilpadre:


CI/SQL/20180920_111931_US14633-Data-Cleanup_HOH-Spouse.sql
```diff
@@ -0,0 +1,65 @@
...
+			DECLARE @Audit_Records dbo.crds_Audit_Item
+
+			UPDATE dbo.Contacts
+			SET Household_Position_ID = 1
+			OUTPUT
+				'Contacts',
+				INSERTED.Contact_ID,
+				'Updated',
+				'Household_Position_ID',
+				'Household_Position_ID',
```
chris-branch: I think this is the Field Label (in the UI), which should be `Household Position`

PR #4960, opened 2018-09-28
US14869 - remove giving tab

jpc11623:
Removed giving tab code and broken unit tests


timmyg:
is this removing the giving tab on profile and the giving history page?

PR #5004, opened 2018-11-19
switch to content block

tatelucas:


bzmillerboy:
Looks good. I confirmed the content block exist in PROD.

PR #5058, opened 2019-01-09
Added a script to create dbo.cr_QA_List_Of_Ids data type

candrews55:


CI/TestSQL/StandardScripts/00.TeardownStoredProcs/CreateDataTypeForCleanUp.sql
```diff
@@ -0,0 +1,13 @@
...
+-- Create User-defined Table Type 
+-- for automation clean up
+-- ================================
+USE [MinistryPlatform]
+GO
+
+-- Create the data type
+CREATE TYPE dbo.cr_QA_List_Of_Ids AS TABLE 
+(
+	ID INT NOT NULL
```
s-henney: Can you add a second column to this that'll store the name of the table the id can be found in? Otherwise we'll have a list of ids with no context for where they came from.

candrews55: This is a data type not a table. It is used to generate a temp table based on its structure.

s-henney: So long as the table name gives us some reference to what table the ids come from, I'm fine with this.

PR #5093, opened 2019-01-30
Feature/netlify

janssejl:
Defaulting redirectToMaestro (causes page refresh to kick back out to crds-net) when angular can't find the route. Also adding a check for the angular home page because that IS found but we never want it to go there.

### DIFF
--- a/crossroads.net/app/routes.content.js
+++ b/crossroads.net/app/routes.content.js
@@ -67,7 +67,7 @@
 
               var childPromise = promise.then(function (originalPromise) {
 
-                if (originalPromise.pages.length > 0) {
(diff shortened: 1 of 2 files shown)


crossroads.net/app/routes.content.js
```diff
@@ -88,44 +88,7 @@
...
-                  } else {
-                    ContentPageService.page = {
-                      content: '404 Content not found',
-                      pageType: '',
-                      title: 'Page not found'
-                    };
-                  }
-                });
-                return notFoundPromise;
+                redirectToMaestro();
```
dillon-courts: Change that function name you heathen.

PR #5104, opened 2019-02-04
remove the smith family from teardown scripts

candrews55:
Smith family is set up in prod. No need to recreate them in demo and int
after a DB refresh.


candrews55:
@s-henney  While that is a risk. The issues I have found have been around data needing to update birth dates and groups. All which needed to be done manually. For example when one of the children needs to be under 2. 

Even if there is a need to recreate the family. The information needed is listed kidsClubGeorgeSmithFamily.cs .

PR #5123, opened 2019-02-11
fixing impersonate for profile

dillon-courts:


dillon-courts:
We need to get the profile of the contactId that is in the parameter. Not the profile of the user associated with the accessToken.

PR #5148, opened 2019-02-21
Feature/us15821 auth updates trips

danrye:


Gateway/crds-angular.test/Services/TripServiceTest.cs
```diff
@@ -985,6 +989,17 @@ private List<MpTripDistribution> MockTripScholarshipDonationsResponse()
             };
         }
 
+        private AuthDTO fakeAuthDTO(int contactId, string emailAddress)
```
andycanterbury: This seems like a useful thing in testing other services also. Would I make sense to move this to a shared place like the Helpers folder or something?

vchtelmakh: Maybe here: https://github.com/crdschurch/crds-fred/blob/development/CrdsFred.Tests/Helpers/FakeModels.cs

danrye: Good call, I had this in the previous PR also. I'd like to do this as part of my next one (I'm currently working childcare) and I'll go back and update trips, payments, and camps

PR #5151, opened 2019-02-21
Refactoring childcare for the new auth

danrye:
- also added a test helper for camps and payments


Gateway/crds-angular.test/Models/Crossroads/AuthHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+            token.UserInfo = new UserInfo();
+            token.UserInfo.Mp = new MpUserInfo();
+            token.UserInfo.Mp.ContactId = contactId;
+            token.UserInfo.Mp.HouseholdId = new int?(householdId);
+            token.UserInfo.Mp.Email = emailAddress;
+
+            return token;
+        }
+    }
+}
```
vchtelmakh: 👍

PR #5199, opened 2019-04-09
[2000's IP] Logzio implementation

j-brammer:
Added logzio configuration everywhere log4net XML configurations were availabl.


j-brammer:
> Will log4net still log to the file on the web server? From what I can tell you didn't change anything but I'm just confirming.

Yes. This is an addition to the config that tells log4net to also send logs to our logzio implementation

PR #5200, opened 2019-04-12
Waivers updated to new Auth

danrye:


Gateway/crds-angular.test/controllers/WaiverControllerTest.cs
```diff
@@ -55,8 +55,7 @@ public void Teardown()
             _userImpersonationService.VerifyAll();
             _authenticationService.VerifyAll();
         }
-
-        [Test]
```
vchtelmakh: Why is the `[Test]` attribute being removed?

danrye: @vchtelmakh because with the new auth the controllers can't be tested. Rather then delete the tests I'm removing the attributes in hopes the new auth becomes testable in the future. So far all controller tests have had to have this done.

PR #5217, opened 2019-05-10
Getting Stuff To Work™️

lukemoderwell:
Changing `crds-angular` to consume the new system-pages.json formatting and fixing a few metadata problems


crossroads.net/app/routes.content.js
```diff
@@ -39,17 +39,14 @@
...
-
-              if(Session.isActive() && link === "/" ) {
-                link = getPersonalizedContentPath(link);
-              }
-
+                
               link = addTrailingSlashIfNecessary(link);
 
-              function redirectToMaestro() {
+              function redirectToCrdsNet() {
```
dillon-courts: I think we should rename this to something like 'redirectOutsideSPA'. You can probably think of a better name but I think the 'redirectToMaestro' name was a poor choice initially as what we are really doing is a window location replace.

crossroads.net/app/routes.content.js
```diff
@@ -148,7 +145,10 @@
                   var firstSentence = content.match(/[^.]*/)[0] + '.';
                   metaDescription = firstSentence;
                 }
-
+                
+                console.log(
```
dillon-courts: Do you want this here still?

PR #5241, opened 2019-06-03
defect/DE6732-infinite-loops

janssejl:


j-brammer:
After doing a little research it seems like it's safe to gitignore this file and it's an 'optional' dependency tracking tool introduced in npm 5.

https://www.codementor.io/johnkennedy/get-rid-of-that-npm-package-lock-json-e0bj7ai42

crossroads.net/package-lock.json
```diff
@@ -15,12 +15,12 @@
       "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
```
j-brammer: If we are going to add this file to .gitignore, you'll probably want to also do a `git rm crossroads.net/package-lock.json` or `git rm package-lock.json` from the crossroads.net directory.

PR #5275, opened 2019-07-09
defect/DE0000-duplicate-user

vchtelmakh:
Prod was returning an error stating a duplicate user exists so the password cannot be reset.

We found that the code was using the old `,` syntax depending on view column order. Changed the code to just look for a specific field.


Gateway/MinistryPlatform.Translation/Repositories/UserRepository.cs
```diff
@@ -2,209 +2,231 @@
...
     }
+
+    private static void ThrowExceptionIfUserDoesNotExistOrIsDupe(string email, List<JObject> usersWithMatchingEmail)
+    {
+      if (!usersWithMatchingEmail.Any())
+      {
+        throw new ApplicationException($"No users with {email} username found");
+      }
+
+      if (usersWithMatchingEmail.Count != 1)
```
dillon-courts: I would probably change this to check for more than 1 user and modify the error message to alert that more than one user was found. You are already checking for the 0 case above.

PR #5280, opened 2019-07-24
US17588 | Christian | add meta data display text for sub categories

christiancrawford:


Gateway/crds-angular/Services/FirestoreUpdateService.cs
```diff
@@ -696,6 +696,23 @@ private string BuildGroupAttrString(int groupid)
                 attrString = $"{attrString} {categoryText}";
             }
 
+            // get sub group categories
+            ObjectAttributeTypeDTO subgroupcategory;
+            if (t.TryGetValue(90, out subgroupcategory))
```
briancavanaugh: The constant on this line needs to be 92 for the subcategories

christiancrawford: good call, thanks for the catch!

PR #5306, opened 2019-09-04
Logging headers to help debug

dillon-courts:


Gateway/crds-angular/Controllers/API/LoginController.cs
```diff
@@ -14,12 +14,13 @@
 using Crossroads.Web.Common.Security;
 using MinistryPlatform.Translation.Repositories;
 using MinistryPlatform.Translation.Repositories.Interfaces;
-
```
bretrex-callibrity: What was wrong with this line? 2/10, wasteful

PR #5324, opened 2019-10-17
Added allow mappings to work with the new TimeZone and Locale info co…

candrews55:
…ming back from MP


Gateway/MinistryPlatform.Translation/Service References/PlatformService/Reference.cs
```diff
@@ -617,6 +617,16 @@ public enum DataTypes : int {
...
+
+        /// <summary>
+        ///  Added Manually. These are needed to allow getting email templates through SOAP API
+        /// </summary>
+        [System.Runtime.Serialization.EnumMemberAttribute()]
+        TimeZone = 23,
+
+        [System.Runtime.Serialization.EnumMemberAttribute()]
+        Locale = 24,
+        //-----------------------------------------------------------------------
```
vchtelmakh: Do we really need the magic comment line?

PR #5341, opened 2019-11-15
Feature/us18316 webhook on password change

rdonnelly02:
Updates for Gateway to call the Identity Service to pass webhook to Differential (Web App) endpoint.

*Note : NUnitTestAdapter.2.2.0 added to package and caused regeneration of app.config's.


Gateway/crds-angular/Services/Interfaces/IPersonService.cs
```diff
@@ -10,7 +10,7 @@ namespace crds_angular.Services.Interfaces
 {
     public interface IPersonService
     {
-        void SetProfile(Person person);        
+        void SetProfile(Person person, string userAccessToken);
```
dmiyamasu: @rdonnelly02 @s-henney How confident are you that changing the method signature has not created any regression issues?

s-henney: Confident. @rdonnelly02 updated all calls to this method and it's not used outside this repo (by crds code)

Gateway/crds-angular/Services/PersonService.cs
```diff
@@ -160,5 +160,71 @@ public Person GetLoggedInUserProfile(String token)
...
+                    throw new Exception("Old password did not match profile");
+                }
+                else
+                {
+                    var userUpdateValues = new Dictionary<string, object>();
+                    userUpdateValues["Display_Name"] = $"{person.LastName}, {person.NickName}";
+                    _userRepository.UpdateUser(userUpdateValues);
+
+                    await UpdateOktaEmailAddressIfNeeded(person, userAccessToken);
+                    await UpdateOktaPasswordIfNeeded(person, userAccessToken);
```
s-henney: In the scenario where a user's email and password are both changed at the same time these methods MUST be called in this order or the call to change the password will be using the wrong email. I think.
Could someone confirm that the email change call will finish before the password change call is made?

Whelans90: I did some quick research. From the Microsoft docs "The await operator suspends evaluation of the enclosing async method until the asynchronous operation represented by its operand completes." 

I interpret that to mean that the UpdateUsernameOrPasswordIfNeeded method, that encloses the two awaited methods, will pause until UpdateOktaEmailAddressIfNeeded returns its Task<bool>, then proceed to the UpdateOktaPasswordIfNeeded.

PR #5385, opened 2020-02-18
US18674-Email Event Listing XML Script

shakila-rajaiah:
Made the required changes suggested by John Eversole. Please merge this code as i do not have merge privileges in GIT.

Thanks,
Shakila Rajaiah.

### DIFF
--- a/CI/SQL/2020-02-18_030300_US18674_StoredProc_report_CRDS_Event_Listing_For_CommunityCare_Email_XML_Report.sql
+++ b/CI/SQL/2020-02-18_030300_US18674_StoredProc_report_CRDS_Event_Listing_For_CommunityCare_Email_XML_Report.sql
@@ -474,11 +474,11 @@ SET     @Body = @Title + N'<table width=''100%'' border=''1'' cellspacing=''0'' 
 
 	-- 2/18/2020: Don't hardcode the values, get them from the respective tables
     --SET @ToContactID = 7837576 -- Hello team, don't hard code this as the  contact ID may change for Demo and prod.
-	--SET @ReplyToContact_ID = 1519180, email = updates@crossroads.net
-	--SET @CommUser_ID = 3009216 -- userID of email= updates@crossroads.net
+	--SET @ReplyToContact_ID = 1519180, where C.Last_Name = 'Communications' and C.First_Name = 'Crossroads'
+	--SET @CommUser_ID = 3009216 -- userID of displayName = Communications, Crossroads 
 	SELECT @ToContactID = C.Contact_ID FROM Contacts C WHERE (C.Last_Name = 'EventlistingReports')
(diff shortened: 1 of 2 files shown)


CI/SQL/2020-02-18_030300_US18674_StoredProc_report_CRDS_Event_Listing_For_CommunityCare_Email_XML_Report.sql
```diff
@@ -0,0 +1,511 @@
...
+	DECLARE @FromEmail varchar(256)
+	DECLARE @ToEmail varchar(256)
+	DECLARE @NewCommMsgID int
+	DECLARE @CommUser_ID int
+
+	-- 2/18/2020: Don't hardcode the values, get them from the respective tables
+    --SET @ToContactID = 7837576 -- Hello team, don't hard code this as the  contact ID may change for Demo and prod.
+	--SET @ReplyToContact_ID = 1519180, email = updates@crossroads.net
+	--SET @CommUser_ID = 3009216 -- userID of email= updates@crossroads.net
+	SELECT @ToContactID = C.Contact_ID FROM Contacts C WHERE (C.Last_Name = 'EventlistingReports')
```
briancavanaugh: Since the where clause is selecting records that are using non-unique fields, there is a chance, however unlikely, that multiple records could get returned here.  Instead, consider doing a "SELECT TOP 1..."  Even still this does not guarantee that the correct record will get selected if multiple records with the same last name exist.

This same comment for the 2 lines below this apply there as well.

I feel that the previous lines that had the hard coded IDs better guarantees the records you are looking for since those IDs will not change.  My suggestion is to use your original code.

shakila-rajaiah: Hi Brian,
Thanks for the feed back. I was thinking about adding "SELECT TOP 1..." to the query. I will change it. I am forwarding John Eversole's email to you as it is kind of long. He seems to have some great suggestions as well.

Thanks,
Shakila.

shakila-rajaiah: The changes in https://www.diffchecker.com/t5Dl2Py9

shakila-rajaiah: > Since the where clause is selecting records that are using non-unique fields, there is a chance, however unlikely, that multiple records could get returned here. Instead, consider doing a "SELECT TOP 1..." Even still this does not guarantee that the correct record will get selected if multiple records with the same last name exist.
> 
> This same comment for the 2 lines below this apply there as well.
> 
> I feel that the previous lines that had the hard coded IDs better guarantees the records you are looking for since those IDs will not change. My suggestion is to use your original code.

Fixed the code..

PR #5396, opened 2020-04-30
Added trace attribute to methods for posting profile

Whelans90:
hopefully this provides more details in New Relic for Gateway distributed traces.


Whelans90:
No longer needed updated INT-WEB agent and it is now reporting.

PR #5451, opened 2020-11-02
Feature/us20196 remove event tool

rdonnelly02:
Notes can be found here: https://docs.google.com/spreadsheets/d/1V_4qUfHMR4hXsebuj0QPTgn-G9cPz0WVtxTiJ-BHhIo/edit#gid=164415335


Gateway/crds-angular/Services/Interfaces/IEventService.cs
```diff
@@ -9,9 +9,6 @@ namespace crds_angular.Services.Interfaces
 {
     public interface IEventService
     {
-        bool CreateEventReservation(EventToolDto eventTool);
-        EventToolDto GetEventReservation(int eventId);
-        EventToolDto GetEventRoomDetails(int eventId);
```
Whelans90, line 14: How did you decide which functions were removed and which functions stayed. Did you delete the controller and the interface, then there were build errors that pointed to the missing functions from the interface?

rdonnelly02: I look to see if and where those methods are referred to.  If they are referred to from only within the controller, service, or test of the endpoint we're removing, I see them as safe to delete.  If they're referred to outside of that endpoint/workflow, I leave them.

Whelans90 (approved):

PR #5482, opened 2021-01-13
feature/US20475 - Add api tests for GET /user & framework redesign

s-henney:
- Added tests for GET /user endpoint
- Redesigned the framework for configuring test scenarios (all contained in the CAT folder). This one starts with a fully runnable test scenario and uses mapping and merging to generate permutations, making it possible share asynchronously fetched values between the request and response steps
- Moved the test logic and assertions into the logic so less test code needs to be duplicated in each file


Gateway/crds-angular.api_test/cypress/shared/enums.ts
```diff
@@ -1,4 +1,3 @@
-  export enum ResponseTypes {
-    json = 'json',
-    text = 'text'
-  }
\ No newline at end of file
+export enum Placeholders {
+  assignedInSetup = "This value should be assigned in the setup step. If it isn't, update the setup."
```
Whelans90, line 2: Should this be a part of a README, or does this make error handling easier for a test run?

s-henney: The message text is to help debug while the test author is developing the test, it should never show up as part of an error once the test is fully written. 
Having a placeholder like this lets us fully define what properties are going to be part of a test's request or verified in its response, but indicate that the value cannot be hard coded. The setup function of a TestScenario is where these values would be fetched (ex. auth tokens via API) or generated (ex. pw reset tokens needed by data setup,  request, and response), hence "assignedInSetup".

Whelans90 (approved):

